PR #1, opened 2025-06-23
Initialized front/backend

z3nL:
Made frontend and backend directories
Initialized frontend with default Vite build
Initialized backend with index.js file
Initialized PostgreSQL database with User, Post, and Comment tables through Prisma ORM


z3nL:
Current state of login and register modals, with console outputs on submission
Image 6-24-25 at 3 04 PM
Image 6-24-25 at 3 04 PM

client/src/App.jsx
```diff
@@ -0,0 +1,60 @@
...
+
+    <header>
+      <h1>Sk8rlog</h1>
+    </header>
+
+    <BrowserRouter>
+      <Routes>
+
+        <Route path='/' element={
+          <form className='landingModal' action={(formData) => handleSubmit(formData, "login")}>
```
itumejia: should we extract these into separate components? (something like `LoginModal` and `RegisterModal`)

itumejia (approved):
Looks good, just left a comment. Let's try to have an organized codebase from the beginning by extracting components that make sense to have their own files. Also, as this will be a larger application compared to the assignments you had been working on, it also makes sense to start creating folders (for example, one folder for all frontend files related to newsfeed, one folder for authentication, etc. )

PR #2, opened 2025-06-25
Login and Register Basics

z3nL:
**Key feature**
Implemented user login/registration validation; successful submissions of either now use useNavigate to direct the user to a dummy Home page that will be populated with the planned home feed page content. With this basic flow complete, I can now implement user authentication functionality to control access across the app.

**Additional work done**
Moved handleSubmit to userUtils and renamed it to handleLoginRegister to be more descriptive; created a "main" components folder and placed Home in it

**Issues**
Returning a 401 status code for invalidation on login/register makes the resulting json inaccessible. I changed the code to 200 now to keep it functional but will be looking up a fix to implement Wednesday morning
- **RESOLVED:** After adding error catches in my axios statements, I was able to control how data was handled when axios threw an "error," i.e. 401 codes

Screenshot of invalid login message:
Image 6-24-25 at 7 16 PM

Screenshot of dummy Home page:
Image 6-24-25 at 7 38 PM

PR #3, opened 2025-06-26
More Frontend Skeleton Work

z3nL:
Implemented Home and Profile page skeletons with sample content; CSS Styling still needs to come into play to establish responsiveness across the site

Home Page:
Image 6-25-25 at 6 46 PM

Profile Page, with Clips tab open:
Image 6-25-25 at 6 47 PM

Profile Page, with Blogs tab open:
Image 6-25-25 at 6 47 PM

Sample of a popup that is displayed when the user clicks their profile picture in the top right that prompts them to select user settings or sign out. Clicking the profile picture again disables it:
Image 6-25-25 at 6 47 PM (1)


client/src/main/ProfilePage/Profile.jsx
```diff
@@ -0,0 +1,51 @@
...
+
+import '/src/css/hasSidebar.css'
+
+const Profile = () => {
+    const headerText = `User's Profile`
+    const CLIPS = "clips"
+    const BLOGS = "blogs"
+
+    const [isViewingClips, setIsViewingClips] = useState(true)
+    const [isViewingBlogs, setIsViewingBlogs] = useState(false)
```
chris4307, line 17: Do you need both of these if one is always the opposite of the other? If you are just planning for additional future logic where this isn't the case, then having both makes sense.

client/src/main/ProfilePage/Profile.jsx
```diff
@@ -0,0 +1,51 @@
...
+import Header from '/src/main/Header'
+import Sidebar from '/src/main/Sidebar'
+import ProfileHead from './ProfileHead'
+import ProfilePostsView from './ProfilePostsView'
+import Footer from '/src/main/Footer'
+
+import '/src/css/hasSidebar.css'
+
+const Profile = () => {
+    const headerText = `User's Profile`
```
chris4307, line 12: const HEADER_TEXT = `User's Profile`

client/src/main/ProfilePage/ProfileHead.jsx
```diff
@@ -0,0 +1,26 @@
+
+import skateboard from '/src/assets/skateboard.png'
+
+import '/src/css/profileHead.css'
+
+const ProfileHead = ({ toggleContentView }) => {
+    const CLIPS = "clips"
+    const BLOGS = "blogs"
```
chris4307, line 8: since you have these constants in multiple files, it might be useful to have a dedicated constants file

client/src/utils/userUtils.js
```diff
@@ -2,19 +2,45 @@ const baseUrl = import.meta.env.VITE_API_URL || 'http://localhost:3000'
...
+// Handle login/register submission
+export const handleLoginRegister = (formData, submissionType, setIsSuccessful) => {
+    const username = formData.get("username"), password = formData.get("password")
+
+    submissionType == OPTIONS.LOGIN      && login(username, password, setIsSuccessful)
+    submissionType == OPTIONS.REGISTER   && register(username, password, setIsSuccessful)
+
+    if (submissionType != OPTIONS.LOGIN && submissionType != OPTIONS.REGISTER) {
+        console.error("Invalid type for handleLoginRegister")
+    }
```
chris4307, line 21: can probably put this before the rest of the code in this function, as there's no point running anything else if the type is invalid.

Also can just check the object values
`if (!Object.values(OPTIONS).includes(submissionType)) {
  console.error("Invalid type for handleLoginRegister")
  return
}`

client/src/utils/userUtils.js
```diff
@@ -2,19 +2,45 @@ const baseUrl = import.meta.env.VITE_API_URL || 'http://localhost:3000'
...
 
-export const login = (username, password) => {
-    console.log("Login:", username, password)
-    // stuff will happen here
+// Handle login/register submission
+export const handleLoginRegister = (formData, submissionType, setIsSuccessful) => {
+    const username = formData.get("username"), password = formData.get("password")
+
+    submissionType == OPTIONS.LOGIN      && login(username, password, setIsSuccessful)
+    submissionType == OPTIONS.REGISTER   && register(username, password, setIsSuccessful)
```
chris4307, line 17: this is fine but a switch case might be easier to read

chris4307 (commented):
Looks solid so far! Just left some small comments

itumejia:
Some quick comments regarding the PR itself:
- It seems like this PR includes the changes from https://github.com/Zen-L-MU2025/project-sk8rlog/pull/2 . Maybe something got messed up with the branches, but let's try to keep the PRs isolated, otherwise it is hard to review each PR separately
- This PR was quite large, an idea of how you could have split this one, could be by having one for the home feed UI, and one for the profile feed UI.
- Let's be more specific on the PR naming too. For this one, for example, it could have been "Implement UI for Home Feed and Profile Feed surfaces" 

Also open question - have you thought how you'll handle responsiveness for home feed considering there are 3 columns?

> My working theory was to restructure the home page when the viewport is below a certain width so that the columns turn into a three-tab situation that only displays one column at a time - Zen

PR #4, opened 2025-06-26
Implemented User Sign-In Authentication in Backend Using JWT Tokens

z3nL:
Key functionality: Leveraged JSON web tokens to store user info and authentication token in client-side session storage after a successful login or registration. The client will now leverage a /verify route to confirm access to a desired resource before redirecting to it, otherwise defaulting to a 401 page on failure.

Additional features: As previously suggested and arguably needed, CLIPS, BLOGS, and WEEKDAYS have been merged into a single constants file that all related components reference; handleLoginRegister has been refactored into handleLogin_or_Register and now consists of a switch statement that defaults to a console error to determine which action to take; some of my error catches were stylized as "err" and are now written fully as "error" for my own readability

A successful call to /auth/verify when user accesses Home Page after signing in
Image 6-26-25 at 12 24 PM

Here is an example response of what invoking /auth/verify with a faulty token may look like. As with other routes, the boolean isSuccessful serves as an indicator to take alternative action, in this case redirection to a 401 Unauthorized page.
Image 6-26-25 at 12 29 PM

Helper function verifyAccess is also configured so that invoking the /auth/verify route is avoided altogether if no token exists in the session storage. Here is a preview of the HTML React will render in that situation or if the token is simply invalid:
   ```
     <Route path='/unauthorized' element={<>
          <p>Snaked! <em>(401 Unauthorized)</em></p>
          <p>Unable to verify access, please sign back in.</p>
          <Link to='/'>Go Back</Link>
        </> } />
```

### DIFF
--- a/client/src/utils/userUtils.js
+++ b/client/src/utils/userUtils.js
@@ -2,7 +2,6 @@ const baseUrl = import.meta.env.VITE_API_URL || 'http://localhost:3000'
 
 import axios from 'axios'
 
-
 // Tracks if landing page submission is login or register
 export const OPTIONS = {
     LOGIN : "login",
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: client/src/utils/userUtils.js [hunk 1]
```diff
@@ -2,45 +2,84 @@ const baseUrl = import.meta.env.VITE_API_URL || 'http://localhost:3000'
...
+            break
+
+        case OPTIONS.REGISTER:
+            register(username, password, setIsSuccessful)
+            break
 
-    if (submissionType != OPTIONS.LOGIN && submissionType != OPTIONS.REGISTER) {
-        console.error("Invalid type for handleLoginRegister")
+        default:
+            console.error("Invalid type for handleLoginRegister")
```
chris4307, line 25: nice!

client/src/utils/userUtils.js
```diff
@@ -2,45 +2,84 @@ const baseUrl = import.meta.env.VITE_API_URL || 'http://localhost:3000'
...
+
+    axios.get(`${baseUrl}/auth/verify`, {headers : { 'Authorization' : `Bearer ${token}` }})
+        .then(res => {
+            const isSuccessful = res.data.isSuccessful
+            if (isSuccessful) {
+                setHasAccess(true)
+            }
+            else {
+                setHasAccess(false)
+            }
```
chris4307:
```suggestion
            setHasAccess(res.data.isSuccessful)
```

server/routes/auth.js
```diff
@@ -0,0 +1,28 @@
...
+
+const webtoken = require('jsonwebtoken')
+const TOKEN_SECRET = process.env.TOKEN_SECRET
+
+const STATUS_CODES = require('../statusCodes')
+
+router.use('/verify', async (req, res) => {
+    try {
+        // Extract the JWT token from the request header
+        const token = req.headers.authorization.split(' ')[1]
```
chris4307, line 12: is there a better way to do this? I'd imagine there's some util function available to get the token from the header without the splitting and hardcoded array index

server/routes/auth.js
```diff
@@ -0,0 +1,28 @@
...
+
+        // Verify the token
+        webtoken.verify(token, TOKEN_SECRET, (error, user) => {
+            if (error) {
+                return res.status(STATUS_CODES.UNAUTHORIZED).json({ message: error.message, isSuccessful: false })
+            }
+            else {
```
chris4307: nit: `else` is unnecessary here

server/routes/auth.js
z3nL: I tried to look into this for a few minutes but came up empty; I added a TODO to come back to this at a later point but for now it should work fine since the auth header's content should always be `Bearer some_token`

PR #5, opened 2025-06-27
Implemented Basic Clientside User Authentication Flow to Access Home and Profile Pages

z3nL:
### Preface
Relevant commit: [implemented working authentication flow in client](https://github.com/Zen-L-MU2025/project-sk8rlog/pull/5/commits/07f3683d4dad8666121f0a61a51d97e344853fe4)

### Overview
Following the [implementation of sign-in authentication in the backend](https://github.com/Zen-L-MU2025/project-sk8rlog/pull/4), the Home and Profile components of the React application now include functionality to account for whether the user trying to access the resource is signed in or not. If the latter occurs, they are redirected to a 401 Unauthorized page that ultimately regresses them to the root of the site to log in again.

### Additional Features
Implemented additional CSS styling and cleaned up constant references in accordance with previous suggestions. Asset references were overhauled to be absolute file paths from /src unless it is being accessed from the same folder

### Test Plan
Home and Profile now implement the displayed logic where a one-time useEffect gives the state hasAccess a meaningful value and a subsequent useEffect tracking hasAccess triggers a redirect if it is false.
Image 6-26-25 at 11 11 PM

The /unauthorized route on display
Image 6-26-25 at 11 15 PM

### DIFF
--- a/client/src/main/HomePage/Home.jsx
+++ b/client/src/main/HomePage/Home.jsx
@@ -1,3 +1,5 @@
+import { useState, useEffect } from 'react'
+import { useNavigate } from 'react-router'
 
 import Header from '/src/main/Header'
 import Home_News from './Home_News'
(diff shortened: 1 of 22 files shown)


z3nL:
### Additional Comments/Concerns
- Since Home and Profile contain identical logic, I'm thinking of housing it in a single Context that both components can reference alongside the signed in user's ID to streamline accessing content specific to them. Would this be a good idea?

- Should I have housed the redirect logic implemented in these components at a higher level so that they don't needlessly render in the first place? I'm imagining this could be better practice to implement in the codebase.

client/src/main/HomePage/Home.jsx
```diff
@@ -1,19 +1,34 @@
...
 
 import Header from '/src/main/Header'
 import Home_News from './Home_News'
 import Home_Clips from './Home_Clips'
 import Home_Blogs from './Home_Blogs'
 import Footer from '/src/main/Footer'
 
-import { WEEKDAYS } from '/src//utils/weekdays'
+import { WEEKDAYS } from '/src/utils/constants'
+import * as user from '/src/utils/userUtils'
```
itumejia, line 11: We should rename the file to `UserUtils` (starting with upper case) and also import it like that (with same naming) when using it in files. Otherwise `user` can be confused as a variable instead of a functions library that we are using. 

Also, unless you foresee that you'll be using a lot of functions from this file, I'd suggest only importing the function that you are using, in this case seems like you are only using 1 so it makes sense to just import that function.

itumejia (commented):
Looks good, just left one comment

client/src/main/HomePage/Home.jsx
z3nL: Copy that!

[[Gen] Updated userUtils references](https://github.com/Zen-L-MU2025/project-sk8rlog/commit/8f6e37881bba137451a21685fc5087ed67a1529a)

PR #7, opened 2025-06-28
Clips/Blogs Page and Color Scheme Update

z3nL:
### Preface
Relevant commit: [Established Clips/Blogs page and changed color scheme](https://github.com/Zen-L-MU2025/project-sk8rlog/pull/7/commits/e9b78d864b2232fa569d583871b825a11edacd6d)

### Overview
I added a clips and blogs page for users to scroll through uploaded content.

I changed the color scheme and opted for a background that isn't just black; I used an online gradient tool to generate one that I thought looked nice.

### Test Plan

Leveraging previously implemented CSS rules, these pages are already confirmed to be responsive through the browser responsiveness tool.

The Clips page with the gradient background implemented/other coloring changes. Blogs page is identical, barring the header that would indicate otherwise. I added a filtering dropdown at the top as a placeholder for when I upload dummy content to the database.
Image 6-27-25 at 5 03 PM


client/src/main/HomePage/Home_Blogs.jsx
itumejia, line 1: seems like this and Home_Clips components are fairly similar and I don't foresee them being different (most likely you'll just have to change some param when fetching from db). So I'll suggest you to think if it's a good idea to simply have one of them, similar to how you handle the profile view. But I'll let you figure out that decision

itumejia (commented):
looks good

client/src/main/HomePage/Home_Blogs.jsx
z3nL: Agreed, I don't foresee them being different either and think it'll be best to condense the components the same way the Profile page was structured

[[Gen] Streamlined Home components; wiped all references to "News"](https://github.com/Zen-L-MU2025/project-sk8rlog/commit/3bdd41a8c93cb208a7626e47b97be1d0fd74deef)

PR #8, opened 2025-07-01
Session Persistence and Code Cleanup

z3nL:
### Overview
Key feature: Using cookies, sessions in the app now persist past the tab they originated in, improving user experience. The cookies are created in the /login and /register routes, rather than the /verify route it was [previously housed in](https://github.com/Zen-L-MU2025/project-sk8rlog/commit/ed523304eb9370ebc1c676388f69918cc41c479c). The latter is now used solely for verifying tokens as it should and does not create cookies.

Additional features:
- Accounting for feedback provided on previous PR's, some updates were made in the codebase to reflect them and promote a cleaner environment

- A visual bug occurred with the overflow scroll bar on any container that used it having a static black background, which was corrected in the CSS (see commits tagged [Gen])

### Test Plan
Accessing Home (or any protected resource) in a new tab used to default to Unauthorized without cookies, which is now corrected. Note that the scroll bars in the video just show the "thumb" itself, which was not the case beforehand.
https://github.com/user-attachments/assets/44f64aba-9278-4415-94ed-248cf54472b0


client/src/main/Posts.jsx
```diff
@@ -1,41 +1,59 @@
...
+                        <option value='near'>Near You</option>
+                    </select>
+                </form>
+                <div className="posts">
+                {
+                    [...Array(20)].map(i => {
+                        return (
+                            <PostCard key={i} postType={postType} />
+                        )
+                    })
```
chris4307: Could you explain why you're using an array of size 20 here?

server/routes/users.js
```diff
@@ -52,9 +54,10 @@ router.use('/login', async (req, res) => {
         }
 
         const tokenPayload = { userID: user.userID, username: user.username}
-        const token = webtoken.sign(tokenPayload, TOKEN_SECRET, { expiresIn: '1h' })
+        const token = await webtoken.sign(tokenPayload, TOKEN_SECRET, { expiresIn: '1h' })
 
-        return res.status(STATUS_CODES.OK).json({ user, isSuccessful: true, token })
+        res.status(STATUS_CODES.OK).json({ user, isSuccessful: true, token })
+        next()
```
chris4307, line 60: just curious, what's this `next()` doing?

client/src/main/ProfilePage/Profile.jsx
```diff
@@ -8,7 +8,7 @@ import ProfilePostsView from './ProfilePostsView'
 import Footer from '/src/main/Footer'
 
 import { CLIPS } from '/src/utils/constants'
-import * as user from '/src/utils/userUtils'
+import { verifyAccess } from '/src/utils/UserUtils'
```
chris4307, line 11: I like how you're importing just the utils you need!

client/src/main/HomePage/HomePostsView.jsx
```diff
@@ -0,0 +1,29 @@
...
+
+import '/src/css/home_main.css'
+
+const HomePostsView = ({ postType }) => {
+    const postType_lowercase = postType.toLowerCase()
+
+    return(<>
+        <div id={`home_${postType_lowercase}`} className="column" >
+
+            <Link to={`/${postType_lowercase}`}>
```
chris4307, line 13: cool that you're using a dynamic div ids and routes! Any edge cases where the link routes to an invalid route? Just want to make sure that there's error handling for this case.

client/src/main/Posts.jsx
z3nL: Whoops, magic number; it's just an arbitrarily larger number to test the visual overflow

[Resolve magic number in Posts](https://github.com/Zen-L-MU2025/project-sk8rlog/commit/6c5c07ab4aa3832eabdfb6d32d93e4de150ca355)

server/routes/users.js
z3nL: Effectively functions as a return since there is nothing after it; I started adding next() as an intermediary in case I need some continuation of the endpoint in the future but I'll cut down on this in the future during cleanup

client/src/main/HomePage/HomePostsView.jsx
z3nL: I wrote a wildcard into the root that accounts for dead routes: https://github.com/Zen-L-MU2025/project-sk8rlog/blob/6c5c07ab4aa3832eabdfb6d32d93e4de150ca355/client/src/App.jsx#L38

server/routes/users.js
chris4307: > Effectively functions as a return since there is nothing after it; I started adding next() as an intermediary in case I need some continuation of the endpoint in the future but I'll cut down on this in the future during cleanup

Makes sense! Just always want to be aware of these future implementation type things. Might want to put a TODO either for cleanup if not used or just saying how you plan on using it in the future (if it's not used yet.) Just for tracking

client/src/main/Posts.jsx
chris4307: > Whoops, magic number; it's just an arbitrarily larger number to test the visual overflow
> 
> [Resolve magic number in Posts](https://github.com/Zen-L-MU2025/project-sk8rlog/commit/6c5c07ab4aa3832eabdfb6d32d93e4de150ca355)

Ah sounds good!

PR #9, opened 2025-07-01
Implemented Google Cloud Storage // A Post Creation Modal

z3nL:
### Preface
The nature of these key features are categorically distant and should have been in their own PR's, but I mistakenly pushed the post creation modal to this branch. Fortunately, this is mitigated by the fact that each commit in this PR solely corresponds to its relevant feature.

### Overview
Key features: 
- I set up a Google Cloud Storage bucket and tested file uploading through the backend after having to sift through documentation and some tutorials to grasp the required configurations. A successful endpoint test confirmed that the skeleton I created will serve the real file upload feature for the app in the future.
- Honing in on the remainder of the app's required features to start the technical challenges, I created a skeleton for the post creation modal that alternates between requested inputs based on selected post type.

### Test Plan
Endpoint Test, resulting storage object is [visible on the internet](https://storage.googleapis.com/sk8rlog-media/skateboard.png)
image

Post Creation Modal Skeleton
https://github.com/user-attachments/assets/d38a6991-294a-437c-a02b-09a99010be1c


client/src/main/ProfilePage/ProfileHead.jsx
```diff
@@ -1,10 +1,15 @@
...
+import CreatePostModal from '../Modals/CreatePostModal'
 
 import { CLIPS, BLOGS } from '/src/utils/constants'
 import skateboard from '/src/assets/skateboard.png'
 
 import '/src/css/profile.css'
 
 const ProfileHead = ({ setProfileContentView }) => {
+    const [showCreatePostModal, setShowCreatePostModal] = useState(false)
+    const toggleCreatePostModal = () => setShowCreatePostModal(!showCreatePostModal)
```
itumejia, line 12: nice naming + cool to have the toggle function to keep it simple since you are passing it through props

server/utils/GCS.js
```diff
@@ -0,0 +1,26 @@
...
+    try {
+        // Init bucket
+        const { BUCKET_NAME, PROJECT_ID, KEYFILE_NAME : keyFilename } = process.env
+        const storage = new Storage({PROJECT_ID, keyFilename})
+        const bucket = storage.bucket(BUCKET_NAME)
+
+        const res = await bucket.upload(file_path, {
+            destination: destination_name
+        })
+        console.log(res)
```
itumejia, line 16: let's remove console logs

server/utils/GCS.js
```diff
@@ -0,0 +1,26 @@
+// Sample of Google Cloud Storage usage in Sk8rlog
+
+import { Storage } from '@google-cloud/storage'
+
+// Upload file to GCS bucket
+export const UploadFile = async (file_path, destination_name) => {
```
itumejia, line 6: since you are naming a function here, let's start with lower case `uploadFile`

client/src/main/Modals/CreatePostModal.jsx
```diff
@@ -0,0 +1,51 @@
...
+import '/src/css/createPostModal.css'
+
+const CreatePostModal = ({ toggleCreatePostModal }) => {
+    const handleForm = (formData) => {
+
+
+        toggleCreatePostModal()
+    }
+
+    const DEFAULT = 'default'
```
itumejia, line 14: let's move this to constants file?

client/src/main/Modals/CreatePostModal.jsx
```diff
@@ -0,0 +1,51 @@
...
+                </>}
+
+                { postType === BLOGS && <>
+                    <input type='text' name='Title' placeholder='Title (50 character maximum)' maxLength={50} required />
+                    <textarea type='long_text' name='Content' placeholder='Description (300 character maximum)' maxLength={300} required />
+                </>}
+
+
+                { postType !== DEFAULT &&
+                    <button type='submit' onSubmit={(event) => event.preventDefault()}>Upload</button>
```
itumejia, line 44: noob question: what is `preventDefault()` doing here?

itumejia (commented):
LGTM, PRs are looking cleaner now. Just left few minor comments

client/src/main/Modals/CreatePostModal.jsx
z3nL: A form's default submission behavior involves a page reload, which I like to avoid when possible.
However, you asking this question made me curious to verify this is actually the case when I remove the preventDefault(). It looks like the default is already prevented anyway because I specified the action the form will carry out in the opening tag, so looks like it's not needed in that line 👍

PR #10, opened 2025-07-02
Sessions Complete

z3nL:
### Overview
Key features: 
- User sessions now persist past the tab they originated in AND reloading the page, the latter of which was not previously accounted for.
- Protected routes now display information relevant to the signed in user, i.e. in the greeting, profile content
- Protected routes keep up with this by loading user data using the userID cookie in the case that the user data is not already in the session storage. 
- A logout route to manually cease sessions will be implemented in the future pursuant to the capstone user sign in/out requirement.

Additional features:
Cleaned up code accounting for feedback left on previous PR's

### Test Plan
Walkthrough of visible changes
https://github.com/user-attachments/assets/e4c52afd-c91f-4a58-ad07-c983af607f47

### DIFF
--- a/client/src/App.jsx
+++ b/client/src/App.jsx
@@ -1,3 +1,4 @@
+import { useState } from 'react'
 import { Routes, Route, BrowserRouter, Link } from "react-router"
 
 import LoginModal from './auth/LoginModal'
@@ -6,13 +7,17 @@ import Home from './main/HomePage/Home'
 import Profile from './main//ProfilePage/Profile'
 import Posts from './main/Posts'
 
+import UserContext from '/src/utils/UserContext.js'
 import { CLIPS, BLOGS } from '/src/utils/constants'
 
 import './App.css'
 
 function App() {
 
+  const [activeUser, setActiveUser] = useState({})
+
   return (<>
+    <UserContext.Provider value={{ activeUser, setActiveUser }}>
 
     <BrowserRouter>
       <Routes>
@@ -43,6 +48,7 @@ function App() {
       </Routes>
     </BrowserRouter>
 
+    </UserContext.Provider>
   </>)
 }
 
--- a/client/src/main/HomePage/Home.jsx
+++ b/client/src/main/HomePage/Home.jsx
@@ -5,14 +5,18 @@ import Header from '/src/main/Header'
 import HomePostsView from './HomePostsView'
 import Footer from '/src/main/Footer'
 
-import UserContext from '/src/utils/UserContext.js'
+import UserContext from '/src/utils/UserContext'
 import { WEEKDAYS, CLIPS, BLOGS } from '/src/utils/constants'
-import { verifyAccess } from '/src/utils/UserUtils'
+import { verifyAccess, loadUserSession } from '/src/utils/UserUtils'
 
 import '/src/css/home_main.css'
 
 const Home = () => {
-    const { activeUser } = useContext(UserContext)
+    const { activeUser, setActiveUser } = useContext(UserContext)
+    useEffect( () => {
+        const load = async () => { await loadUserSession(setActiveUser) }
+        load()
+    }, [])
 
     const navigate = useNavigate()
 
--- a/client/src/main/HomePage/Home.jsx
+++ b/client/src/main/HomePage/Home.jsx
@@ -1,16 +1,19 @@
-import { useState, useEffect } from 'react'
+import { useState, useEffect, useContext } from 'react'
 import { useNavigate } from 'react-router'
 
 import Header from '/src/main/Header'
 import HomePostsView from './HomePostsView'
 import Footer from '/src/main/Footer'
 
+import UserContext from '/src/utils/UserContext.js'
 import { WEEKDAYS, CLIPS, BLOGS } from '/src/utils/constants'
 import { verifyAccess } from '/src/utils/UserUtils'
 
 import '/src/css/home_main.css'
 
 const Home = () => {
+    const { activeUser } = useContext(UserContext)
+
     const navigate = useNavigate()
 
     const [hasAccess, setHasAccess] = useState(null)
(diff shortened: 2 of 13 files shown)


client/src/main/HomePage/Home.jsx
```diff
@@ -23,7 +30,8 @@ const Home = () => {
         hasAccess === false && navigate('/unauthorized')
     }, [hasAccess])
 
-    const HEADER_TEXT = `Happy ${WEEKDAYS[new Date().getDay()]}, User!`
+    const user_title = activeUser.name ? activeUser.name : `@${activeUser.username}`
```
itumejia, line 33: let's keep naming conventions here (`userTitle`)

client/src/App.jsx
```diff
@@ -6,13 +7,17 @@ import Home from './main/HomePage/Home'
...
 import { CLIPS, BLOGS } from '/src/utils/constants'
 
 import './App.css'
 
 function App() {
 
+  const [activeUser, setActiveUser] = useState({})
+
   return (<>
+    <UserContext.Provider value={{ activeUser, setActiveUser }}>
```
itumejia, line 20: nice!

client/src/main/Modals/CreatePostModal.jsx
```diff
@@ -1,17 +1,25 @@
...
+    const { activeUser, setActiveUser } = useContext(UserContext)
+    useEffect( () => {
+        const load = async () => { await loadUserSession(setActiveUser) }
+        load()
+    }, [])
+
     const handleForm = (formData) => {
 
+        const formObject = Object.fromEntries([...formData])
+        console.log(formObject)
```
itumejia, line 19: remember to remove console logs after you're done debugging

client/src/main/ProfilePage/Profile.jsx
```diff
@@ -26,7 +33,8 @@ const Profile = () => {
         hasAccess === false && navigate('/unauthorized')
     }, [hasAccess])
 
-    const HEADER_TEXT = `User's Profile`
+    const user_title = activeUser.name ? activeUser.name : `@${activeUser.username}`
```
itumejia, line 36: same

client/src/utils/userUtils.js
```diff
@@ -27,15 +29,15 @@ export const handleLoginOrRegister = (formData, submissionType, setIsSuccessful)
 }
 
 // Handle registration: store token and new user in session storage on completion
-export const register = (username, password, setIsSuccessful) => {
-    axios.post(`${baseUrl}/users/register`, {username, password, withCredentials: true})
+export const register = (formObject, setIsSuccessful) => {
```
itumejia, line 32: great to see how you're passing formObject instead of the variables separated

client/src/utils/userUtils.js
```diff
@@ -99,3 +98,41 @@ export const verifyAccess = (setHasAccess) => {
 
 
 }
+
+// Fetch user data to load into session storage
+// Sets activeUser to user data if successful, logs error to console otherwise
+// This function is only called from protected routes under the assumption that user is already logged in
+export const loadUserSession = (setActiveUser) => {
```
itumejia, line 105: cool!

server/routes/auth.js
```diff
@@ -1,3 +1,6 @@
+// TODO next() is a placeholder in case of some continuation of an endpoint in the future
```
itumejia, line 1: this would be more useful if you leave the comment right next to the line you use `next()`

itumejia (commented):
Left some comments. Some comments in the PR structure:
- Test plan: Since you are adding both login and register logic here, your test plan should cover both. So I suggest adding another video with the register feature
- PR title: (more of a stretch). Let's try being more explicit on the PR naming, specially since you are building on top of an authentication flow that already existed, a clearer name would help to know what exactly you are changing/improving, and "Sessions complete" doesn't tell much. You could name it something like "Implement sessions persistence by using browser cookies"

client/src/main/Modals/CreatePostModal.jsx
z3nL: Not to try excusing that a console log was included in the first place, but #11 ultimately omits this in the codebase; I'll make sure console logs are cleared from a PR from now on even if a direct successor takes care of it (excluding any that might already be present in #11 since I already created it, but my working branch that will form #12 does not have any at the moment)

client/src/utils/userUtils.js
z3nL: I'm normally a big fan of having the variables separated so I know exactly what's supposed to be present but the conversion from FormData to a complete Object made working with this portion a lot easier

client/src/main/HomePage/Home.jsx
z3nL: Thank you! In pointing this out you made me realize I never cleaned up the declaration for user title. I've been trying to keep my conditional declarations as simple as possible after Chris once showed me a simpler way to do them if the condition is just an "existence check."

[[Gen] Cleanup pursuant to comments left in](https://github.com/Zen-L-MU2025/project-sk8rlog/commit/a086b960d98a56c51e966ea7e93c87a6805f0468) https://github.com/Zen-L-MU2025/project-sk8rlog/pull/10

server/routes/auth.js
z3nL: Agreed, but I think for readability's sake it can do fine without any mention of it until it's needed
 
[[Gen] Removed "next" entirely](https://github.com/Zen-L-MU2025/project-sk8rlog/commit/e9d2b2ff54410da27708625ec8bd18a3351615d9)

client/src/main/ProfilePage/Profile.jsx
z3nL: [[Gen] Update user title declaration in Profile](https://github.com/Zen-L-MU2025/project-sk8rlog/commit/c6a0d0249a2c1069037957a20ee7e54d678289b4)

PR #11, opened 2025-07-02
Post Creation Feature Complete

z3nL:
### Preface
My original implementation of Google Cloud Storage file uploads worked in the context of handing GCS an absolute file path from my computer, but this behavior can not be replicated with file uploads in an HTML form, which became the primary blocker and focus in this PR.

### Overview
The post creation modal is complete and fully integrated with the backend. 
- Users can now create a Clip/Blog post and upload their video or header photo, respectively
- The file is subsequently uploaded to the app's GCS bucket, after which the post, including the URL to the file in the cloud, is uploaded to the database

### Test Plan
This does not include a video demonstration of a Blog upload along with a Clip upload as it should for well-roundedness. Please assume that it would have functioned the same way at this point, as a subsequent PR will contain a full demonstration of both in action.

Video demonstration of Clip upload
https://www.loom.com/share/bf33d02bdc404920bbb54c22f9b19b02?sid=b74e095f-9322-4327-a4ce-6b274285c4ec


client/src/main/Modals/CreatePostModal.jsx
```diff
@@ -13,14 +14,12 @@ const CreatePostModal = ({ toggleCreatePostModal }) => {
...
     }, [])
 
-    const handleForm = (formData) => {
-
-        const formObject = Object.fromEntries([...formData])
-        console.log(formObject)
+    const [postType, setPostType] = useState(DEFAULT)
 
+    const handleForm = (formData) => {
+        uploadPost(postType, formData, activeUser.userID, activeUser.location)
```
itumejia, line 20: I'm curious, where are we getting the location from?

itumejia (commented):
nice clean code :)

client/src/main/Modals/CreatePostModal.jsx
```diff
@@ -37,14 +36,17 @@ const CreatePostModal = ({ toggleCreatePostModal }) => {
                 </select>
 
                 { postType === CLIPS && <>
-                    <input type='file' id='clip' name='clip' accept='video/*' />
-                    <input type='text' name='description' placeholder='Description (50 character maximum)' maxLength={50} required />
+                    <div className='upload'>
+                        <p className='message'>Select a clip to upload:</p>
+                        <input type='file' id='clip' name='postFile' accept='video/*' required/>
+                    </div>
+                    <input type='text' name='textContent' placeholder='What are we looking at? (50 char. maximum)' maxLength={50} required />
```
stephanieprussick, line 43: Would be a cool follow up to implement the character limit via the UI instead of explicitly saying "(50 char. maximum)" kinda like how it's done on X or similar tools

server/routes/posts.js
```diff
@@ -0,0 +1,56 @@
...
+
+// POST /posts/uploadFile
+// Uploads a pending post's file to GCS and returns the URL
+router.post('/uploadFile', multer.single('postFile'), async (req, res, next) => {
+    try {
+        if (!req.file) {
+            return res.status(STATUS_CODES.BAD_REQUEST).json({ message: 'No file provided', isSuccessful: false })
+        }
+
+        let extension = await new String(req.file.mimetype).split('/')[1]
```
stephanieprussick, line 22: You have a try/catch which is good (so in prod, an exception would be thrown), but with situations where we are checking the specific index of an array like array[1] I like to ensure that exists and then handle the case where it does not

like 

```
if (extension.length < 2) {
  // how should we handle this case? 
}

```

server/routes/users.js
```diff
@@ -69,11 +67,12 @@ router.use('/login', async (req, res, next) => {
 
 // GET /users/:userID
 // Returns the user object for the given userID
-router.use('/:userID', async (req, res, next) => {
+router.get('/:userID', async (req, res, next) => {
     const { userID } = req.params
     const user = await prisma.user.findUnique({ where: { userID } })
     if (!user) {
-        return res.status(STATUS_CODES.NOT_FOUND).json({ message: 'User not found' })
+        console.error(`User with ID ${userID} not found`)
```
stephanieprussick, line 74: In a real prod scenario, you wouldn't want to log someone's userID like this because someone could inspect it and then fake like they are them. This would be a privacy violation at Meta 😝

client/src/main/Modals/CreatePostModal.jsx
z3nL: A User object in the database contains a "location" column, serving as a placeholder for users being able to pick their general location on the map when they create an account and having the Google Geocoding API reverse geocode the latitude/longitude of the selected location into a locality name that would be displayed on the user's profile. 

Broadly, this is in the vein of an idea that matching user locations against post locations could have a role in the recommendation algorithm

client/src/main/Modals/CreatePostModal.jsx
z3nL: I thought so too! I kept note of it as a mini-stretch feature to implement later

server/routes/posts.js
z3nL: Very fair point, I will investigate what happens when extension is not of format "generalType/specific" to determine an appropriate course of action

server/routes/users.js
z3nL: Duly noted - this got me thinking of where else this privacy leak could be taking place in my app, thanks for bringing it up

[[Misc] remove logging userID to console](https://github.com/Zen-L-MU2025/project-sk8rlog/commit/18c9b10dc45963142d28626170abad6da77c05b2)

PR #13, opened 2025-07-03
Logout Button Ends Session and Sends User Back to Login Page

z3nL:
### Overview
The app's header had a Logout button that did not actually do anything up to now. Pursuant to the [Capstone project requirements](https://sites.google.com/meta.com/metau-intern/capstone-project/project-guide), I Implemented its functionality (see PR title).

Key Features
- handleLogout helper function clears session storage and sets all active cookies to empty content with an expiration date set in the past, which forces its instant expiration
- In implementing the logout flow, I mitigated a long-standing issue where logging in would send the user to the home page **before** the session was fully initialized, which would result in ending up at the /unauthorized page even though the user is actually signed in. To fix this, I had to [make the login flow asynchronous](https://github.com/Zen-L-MU2025/project-sk8rlog/commit/cccf9bc27db6c5c9e3f3720fae76f0de48419072), which I originally overlooked

### Test Plan
[Video demonstration](https://www.loom.com/share/5360bf8ed91947318be9f93bd6810961?sid=60ac9449-5a0c-4541-9cb2-433183095448) of user logging out, being taken back to the login page, and no longer being able to access protected routes.
- Note that the end of the video shows the user signing back in and successfully landing on the home page without error, which was not explicitly defined behavior up to the login flow fix.

### DIFF
--- a/client/src/main/Header.jsx
+++ b/client/src/main/Header.jsx
@@ -13,6 +13,15 @@ const Header = ({ HEADER_TEXT }) => {
         setIsIconOverlayOpen(!isIconOverlayOpen)
     }
 
+    const handleLogout = () => {
+        sessionStorage.clear()
+        const cookies = document.cookie.split(';')
+        cookies.forEach(cookie => {
+            // Arbitrarily older date that will force the cookie to instantenously expire
+            document.cookie = cookie.split('=')[0] + '=;expires=Thu, 01 Jan 1984; path=/'
+        })
+    }
+
     return (
         <header className="mainHeader">
             <Link to='/home'> <img className='logo' src={skateboard} alt="skateboard" /> </Link>
@@ -27,7 +36,7 @@ const Header = ({ HEADER_TEXT }) => {
                 <div className='iconOverlay'>
                     <img className='gearIcon' onClick={toggleIconOverlay} src={gear} alt='gear icon' />
                     <button>Settings</button>
-                    <button>Log Out</button>
+                    <Link to='/'><button onClick={handleLogout}>Log Out</button></Link>
                 </div>
             }
         </header>
--- a/client/src/utils/userUtils.js
+++ b/client/src/utils/userUtils.js
@@ -10,17 +10,17 @@ const OPTIONS = {
 }
 
 // Handle login/register submission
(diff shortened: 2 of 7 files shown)


client/src/utils/userUtils.js
```diff
@@ -10,17 +10,17 @@ const OPTIONS = {
...
 // Handle login/register submission
-export const handleLoginOrRegister = (formData, submissionType, setIsSuccessful) => {
+export const handleLoginOrRegister = async (formData, submissionType, setIsSuccessful) => {
 
     const formObject = Object.fromEntries([...formData])
 
     switch (submissionType) {
         case OPTIONS.LOGIN:
-            login(formObject, setIsSuccessful)
+            await login(formObject, setIsSuccessful)
```
itumejia, line 19: I'm curious why did we have to change these to await. I mean it makes sense for them to be asynchronous but wondering how it was working before

client/src/main/Header.jsx
```diff
@@ -13,6 +13,15 @@ const Header = ({ HEADER_TEXT }) => {
         setIsIconOverlayOpen(!isIconOverlayOpen)
     }
 
+    const handleLogout = () => {
+        sessionStorage.clear()
+        const cookies = document.cookie.split(';')
+        cookies.forEach(cookie => {
+            // Arbitrarily older date that will force the cookie to instantenously expire
+            document.cookie = cookie.split('=')[0] + '=;expires=Thu, 01 Jan 1984; path=/'
```
stephanieprussick, line 21: this seems a little hacky
is there a better way to do this?

z3nL: I was thinking about this and want to look into it, "hacky" is the perfect way to put it

client/src/utils/userUtils.js
z3nL: Without the await, the Login/Registration modals would proceed to the home page before actually completing the login / registration flow, which includes creating cookies and initializing session storage. Loading the home page without this completed defaults to /unauthorized without having any information to validate the session. 

As for _why_ it proceeds to the home page too early even though handleLoginOrRegister is called with await, I'm honestly at a loss

PR #14, opened 2025-07-07
Individual Post View // Styling: Fade-in Transition and Gear Icon Hover Easter Egg

z3nL:
### Preface
I wanted to try something new here and combine both the styling and functionality into one PR because they both largely constitute the feature as a whole, let me know if you guys think this would've been better separate 👍 

### Overview - Key Features
A) Users can now click on a post's card in the feed and will be redirected to a page containing the post itself and more information about it. A "Go Back" button at the bottom allows them to return to the page they were redirected from in the interest of user experience.

B) The Single Post, Clips/Blogs, and Profile pages now all share a fade-in transition to add flair to the display while content is loading. It is worth nothing that once you are on Clips or Blogs, directly switching to the opposite does not trigger this effect.

C) Pursuant to creating "an interesting cursor interaction" (guidelines), hovering the gear icon in the header and the Setting/Log Out buttons that show up on click change the cursor to the app's skateboard logo, with the top right of the image acting as the point of contact.

### Test Plan
A/B) [Video](https://www.loom.com/share/c0ca942d0cee4dcab55415bd46c1e42c?sid=9d9a7618-5c7a-4b99-b051-de10b0a9d3a8) of loading a single post with demonstration of Go Back functionality. Note the fade-in transition in action for the pages they are applied to

C) [Video](https://www.loom.com/share/01ea11e2cbb94034b259d61b16bc877f?sid=dad3d293-9733-480c-b22e-92748730b146) of a sample of the implemented custom cursor

### DIFF
--- a/client/src/main/HomePage/Home.jsx
+++ b/client/src/main/HomePage/Home.jsx
@@ -36,7 +36,7 @@ const Home = () => {
     const userTitle = activeUser.name || `@${activeUser.username}`
     const HEADER_TEXT = `Happy ${WEEKDAYS[new Date().getDay()]}, ${userTitle} !`
 
-    if (!isReadyToDisplayContent) return (<p>Loading profile...</p>)
+    if (!isReadyToDisplayContent) return (<p>Loading home page...</p>)
 
     return (<>
         <Header HEADER_TEXT={HEADER_TEXT} />
--- a/client/src/main/HomePage/Home.jsx
+++ b/client/src/main/HomePage/Home.jsx
@@ -38,7 +38,7 @@ const Home = () => {
 
     if (!isReadyToDisplayContent) return (<p>Loading profile...</p>)
 
-    if (isReadyToDisplayContent) return (<>
+    return (<>
         <Header HEADER_TEXT={HEADER_TEXT} />
 
         <h2 className='subHeading'>Today in Sk8rlog</h2>
--- a/client/src/main/HomePage/Home.jsx
+++ b/client/src/main/HomePage/Home.jsx
@@ -14,11 +14,11 @@ import '/src/css/home.css'
 const Home = () => {
     const { activeUser, setActiveUser } = useContext(UserContext)
 
-    const[ isReady, setIsReady ] = useState(false)
+    const[ isReadyToDisplayContent, setIsReadyToDisplayContent ] = useState(false)
     useEffect( () => {
         const loadUser = async () => { await loadUserSession(setActiveUser) }
         loadUser()
-        setIsReady(true)
+        setIsReadyToDisplayContent(true)
     }, [])
 
     const navigate = useNavigate()
@@ -36,9 +36,9 @@ const Home = () => {
     const userTitle = activeUser.name || `@${activeUser.username}`
     const HEADER_TEXT = `Happy ${WEEKDAYS[new Date().getDay()]}, ${userTitle} !`
 
-    if (!isReady) return (<p>Loading profile...</p>)
+    if (!isReadyToDisplayContent) return (<p>Loading profile...</p>)
 
-    if (isReady) return (<>
+    if (isReadyToDisplayContent) return (<>
         <Header HEADER_TEXT={HEADER_TEXT} />
 
         <h2 className='subHeading'>Today in Sk8rlog</h2>
--- a/client/src/main/HomePage/HomePostsView.jsx
+++ b/client/src/main/HomePage/HomePostsView.jsx
@@ -5,6 +5,7 @@ import PostCard from '/src/main/PostCard'
 
 import { HOME_PAGE_POST_COUNT } from '/src/utils/constants'
 import  { getAllPostsByType } from '/src/utils/postUtils'
+import { HOME } from '/src/utils/constants'
 
 import '/src/css/home.css'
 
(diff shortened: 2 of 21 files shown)


client/src/main/SinglePost.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+    return (<>
+        <Header HEADER_TEXT={HEADER_TEXT}/>
+        <div className='singlePostContent'>
+            <p>{ toSingular(post?.type) } posted by <em>@{ postAuthor?.username }</em> on { post?.creationDate }</p>
+            { post?.type === CLIPS && <video src={post?.fileURL} controls={true} className='singlePostMedia'/>}
+            { post?.type === BLOGS && <img src={post?.fileURL} className='singlePostMedia' /> }
+            <p>📍 {post?.location}</p>
+            <p>{post?.description}</p>
+            <p>{post?.likeCount} likes</p>
```
z3nL, line 34: Caught on to "?" syntax to only reference a property iff the object exists and wanted to try it out. I think it turned out well, but I'm wondering if this is generally sensible usage - do you guys think it would be better to wrap the relevant parts of this component in an `if (post) {// functionality} else {// handle no content error}`?

client/src/main/SinglePost.jsx
```diff
@@ -0,0 +1,41 @@
...
+import Footer from './Footer'
+
+import { getUserByID } from '/src/utils/userUtils'
+import { getPostByID } from '/src/utils/postUtils'
+import { CLIPS, BLOGS, toSingular, ORIGINS } from '/src/utils/constants'
+
+import '/src/css/singlePost.css'
+
+const SinglePost = () => {
+    const { origin, postID } = useParams()
```
z3nL, line 14: I noticed that when working on [the Profile page](https://github.com/Zen-L-MU2025/project-sk8rlog/blob/716749a11efd43ee7b103da8f0e30ee5dfb515de/client/src/main/ProfilePage/Profile.jsx#L19), I would try to group related state/effect hooks together. However, I think this looks too scattered and I can only follow it because I wrote it. I tried to change that in SinglePost and group all the state hooks together directly followed by the effect hooks. Do you guys think this approach to hook organization works better or is grouping related hooks (and making it readable) more ideal?

client/src/main/HomePage/Home.jsx
```diff
@@ -14,11 +14,11 @@ import '/src/css/home.css'
 const Home = () => {
     const { activeUser, setActiveUser } = useContext(UserContext)
 
-    const[ isReady, setIsReady ] = useState(false)
+    const[ isReadyToDisplayContent, setIsReadyToDisplayContent ] = useState(false)
```
chris4307, line 17: nice!

client/src/main/HomePage/HomePostsView.jsx
```diff
@@ -3,20 +3,19 @@ import { Link } from 'react-router';
...
 import PostCard from '/src/main/PostCard'
 
+import { HOME_PAGE_POST_COUNT } from '/src/utils/constants'
 import  { getAllPostsByType } from '/src/utils/postUtils'
+import { HOME_ORIGIN } from '/src/utils/constants'
 
 import '/src/css/home.css'
 
-const HomePostsView = ({ activeUser, postType }) => {
+const HomePostsView = ({ postType }) => {
```
chris4307, line 12: great job removing dead variables

client/src/main/HomePage/HomePostsView.jsx
```diff
@@ -28,9 +27,9 @@ const HomePostsView = ({ activeUser, postType }) => {
 
             <div id={`${postType_lowercase}ColumnContent`} className="columnContent" >
                 {
-                    posts.slice(0, 4).map(post => {
+                    posts.slice(0, HOME_PAGE_POST_COUNT).map(post => {
```
chris4307, line 30: nice

client/src/main/PostCard.jsx
```diff
@@ -15,24 +16,25 @@ const PostCard = ({ post, postType }) => {
...
+        return
+    }
 
     return (
+        <Link to={`/${origin}/post/${post.postID}`} className='postRedirect'>
         <article className={`${postType}PostCard`}>
             { postType === CLIPS &&
                 <video ref={embedRef} className={`${postType}PostCardEmbed`} src={post.fileURL}
-                    onMouseEnter={onEnter} onMouseLeave={onLeave}
+                    muted onMouseEnter={onEnter} onMouseLeave={onLeave}
```
chris4307, line 29: noobq what is muted doing here?

chris4307 (commented):
Love the fun features. Not a ton of comments from me today. Great job!

client/src/main/PostCard.jsx
z3nL: Preview video on card will not include audio on hover because that seemed obnoxious to me idk

z3nL (commented):
Requesting Feedback

client/src/main/SinglePost.jsx
itumejia: nope, definitely better practice to use '?' as it leaves the code in a cleaner state :)

client/src/main/SinglePost.jsx
itumejia: yeah there's not a definite rule here, but it's usually easier to read if you put all state hooks first and all useEffect hooks afterwards.

PR #15, opened 2025-07-08
[Posts] Users Can Delete Posts From Their Profile

z3nL:
### Overview
Key Feature
As title states, a user can delete a post from their profile page. The post's content is subsequently deleted from the database and the media file is removed from the Google Cloud Storage bucket.

Additionals
The "Settings" button in the header was meant to allow the user to edit their profile information. I don't think this is crucial to the app's complexity and is not exactly challenging, so I'm slating it to be implemented later on and I removed it from the display for now.

### Test Plan
[Sample](https://www.loom.com/share/b219d50172424e7a99aef51e4867db20?sid=b26e9c6a-9ebd-47fb-a633-3dc853964c1f) of a user deleting a Blog from their profile. The contents are visibly wiped from the database and GCS bucket.

[Sample](https://www.loom.com/share/8e8ae7ab3378400292f32535b83a0096?sid=d0f75c59-fd10-43d7-82a5-622edef0ac89) of a user deleting a Clip from their profile. The same applies.

Screenshot of app header with gear icon toggle open (settings button is now gone)
image


client/src/utils/PostUtils.js
```diff
@@ -58,3 +58,22 @@ export const getPostByID = async ( postID, setPost ) => {
             console.error("getPostByID error: ", error)
         })
 }
+
+// Provided a postID, deletes the post and updates the user's posts array state
+export const deletePost = async ( post, setUserPosts ) => {
```
itumejia, line 63: you don't have to do this change, but just as an fyi, with this type of logic where you have a chain of events occuring in the back-end (deleting the post, and then deleting the file), it's a better practice to handle all this logic from the back-end instead of making 2 separate calls from the front-end. This reduces room for errors when referencing these api endpoints

itumejia (commented):
👍

PR #16, opened 2025-07-09
[TC1][Liking, Tokenization] User can like posts and content will be tokenized + saved

z3nL:
### Overview
- Users can now (un)like a post and this is reflected visually with a heart icon changing between empty/full
- When a user (un)likes a post, its count updates accordingly. This is reflected in a post's individual view
- Liked posts are saved in the database and subsequently persist across page reloads / sessions
- When a user (un)likes a post, the post's content is tokenized and accounted for in the user's _user_Frequency_ object (see [TC1 Documentation](https://docs.google.com/document/d/1UAnIKkqvQI-ZIUi9KWeNQttyQ6jO09DCiWzuDfIl2ss/edit?tab=t.0#heading=h.3qjamfm2ni3r))

### Test Plan
[Video demonstration](https://www.loom.com/share/b002df7654de4b80a9eff3238bfa14f1?sid=7e644089-ee7a-43bf-a88e-f3cae5b6122e) of user liking a post and its status persisting across pages/reload
- Note like count changing on individual post view

[Video demonstration](https://www.loom.com/share/81691011c9104ca69e7c55809ecdfc23?sid=f57f1b09-16b1-4cd1-a602-616072532dbb) of user's _user_Frequency_ object being updated on both like and unlike


client/src/utils/tokenization.js
```diff
@@ -0,0 +1,38 @@
...
+        // Further filter <3 character tokens
+        if (token.length > 2) {
+
+            if (!user_Frequency[token]) {
+                user_Frequency[token] = {}
+            }
+
+            user_Frequency[token]["totalFrequencyAcrossLikedPosts"] = (user_Frequency[token]["totalFrequencyAcrossLikedPosts"] || 0) + frequencyFactor
+
+            if (!tokensUsedInThisInstance.includes(token)) {
```
itumejia: code is getting complex here, so a couple comments here would help to easily understand what you are doing. As general guidance, whenever you think code may be hard to read for folks that don't have much context, then it is a good idea to leave a small comment to help us understand

client/src/utils/tokenization.js
```diff
@@ -0,0 +1,38 @@
...
+
+    const contentToArray = content.split(/[^a-zA-Z0-9]/)
+
+    const filteredContent = removeStopwords(contentToArray)
+
+    const tokensUsedInThisInstance = []
+
+    filteredContent.forEach(token => {
+        // Further filter <3 character tokens
+        if (token.length > 2) {
```
itumejia: Generally you should try to avoid writing as less nested 'if' statements as possible, as it makes code harder to read. In this case, I would suggest first filtering tokens < 3 char, and then going through that new array to normalize it. That way you'll avoid having everything inside the first 'if' statement

itumejia (commented):
Pretty cool to see you start making progress on your first TC! Minor feedback for the PR, I think this one could have easily been broken down into 2 PRs: one for front-end implementation + updating likes counter in the server side, and another one to tokenize and update the userFrequency object

client/src/utils/tokenization.js
z3nL: Ironically enough, this is loosely a callback to a [question I once asked](https://github.com/Zen-L-MU2025/project-sk8rlog/pull/14#discussion_r2190679876). Now that I made the error I see how much better it looks after I move it out

[Added comments, moved token length filter out of iterator](https://github.com/Zen-L-MU2025/project-sk8rlog/pull/16/commits/f0261f88e300c5ec4b81c9c9ee875f090f3554b4)

client/src/utils/tokenization.js
z3nL: Agreed

[Added comments, moved token length filter out of iterator](https://github.com/Zen-L-MU2025/project-sk8rlog/pull/16/commits/f0261f88e300c5ec4b81c9c9ee875f090f3554b4)

client/src/main/SinglePost.jsx
```diff
@@ -22,9 +27,26 @@ const SinglePost = () => {
...
+            case UNLIKE:
+                setPostLikeCount(prev => prev-1)
+                break
+
+            default:
+                console.error('Invalid action')
+        }
+        handleLikeOrUnlikePost(event, post, action, activeUser, setActiveUser)
+    }
+
```
chris4307, line 49: should we move this into utils?

client/src/utils/PostUtils.js
```diff
@@ -77,3 +79,71 @@ export const deletePost = async ( post, setUserPosts ) => {
...
+}
+
+// add postID to user's likedPosts array, update user's frequency object, increment post's like count
+const likePost = async ( post, activeUser, setActiveUser ) => {
+    const postID = post.postID
+    const updatedUserFrequency = await tokenize(post, activeUser, LIKE)
+
+    const updatedUser = {
+        ...activeUser,
+        likedPosts: activeUser.likedPosts ? [...activeUser.likedPosts, postID] : [postID],
```
chris4307, line 108: nit: is there a reason we're using an array here? Is there anything stopping us from using something like a Set? should more time efficient for add/delete/lookups vs an array.

If all we need to do is check existence, add, and remove, it might be a good option.

If you want to store more complicated metadata (timestamp, etc.), then it's possible Set isn't the way to go. Just depends on what you expect this data structure to store/do

client/src/utils/PostUtils.js
```diff
@@ -77,3 +79,71 @@ export const deletePost = async ( post, setUserPosts ) => {
...
+
+    await axios.put(`${baseUrl}/users/${activeUser.userID}/likedPosts/like`, { postID, updatedUserFrequency })
+        .catch(error => {
+            console.error("handleLikeOrUnlikePost error: ", error)
+        })
+
+    await axios.put(`${baseUrl}/posts/${postID}/likes/increment`)
+        .catch(error => {
+            console.error("handleLikeOrUnlikePost error: ", error)
+        })
```
chris4307, line 123: just curious, what happens if the first call succeeds, and the second fails? Also, is it possible to run these concurrently here?

client/src/utils/PostUtils.js
```diff
@@ -77,3 +79,71 @@ export const deletePost = async ( post, setUserPosts ) => {
...
+
+    await axios.put(`${baseUrl}/users/${activeUser.userID}/likedPosts/unlike`, { postID, updatedUserFrequency })
+        .catch(error => {
+            console.error("handleLikeOrUnlikePost error: ", error)
+        })
+
+    await axios.put(`${baseUrl}/posts/${postID}/likes/decrement`)
+        .catch(error => {
+            console.error("handleLikeOrUnlikePost error: ", error)
+        })
```
chris4307, line 148: same question here

client/src/utils/tokenization.js
```diff
@@ -0,0 +1,40 @@
...
+// Tokenize the content of a post, remove stop words
+// If the user has liked the post, increment the frequency of the tokens in the post; decrement if user is unliking
+// Store updated frequency in user.user_Frequency
+export const tokenize = async (post, activeUser, action) => {
+    const content = post.description
+    const user_Frequency = activeUser.user_Frequency || {}
+
+    const frequencyFactor = action === LIKE ? 1 : -1
+
+    const contentToArray = content.split(/[^a-zA-Z0-9]/)
```
chris4307, line 13: maybe put this regex in constants with a descriptive name?

chris4307 (commented):
This looks great! Just a couple small comments

client/src/main/SinglePost.jsx
z3nL: I'd rather not but I take it you asked because of its size; the error handling here is a little redundant because _action_ won't ever be anything other than LIKE or UNLIKE (checked all calls to it to verify)

[[misc] condense handleHeartClick](https://github.com/Zen-L-MU2025/project-sk8rlog/commit/7831b404cb050099f3b5ab12211ad8447e3e7159)

client/src/utils/PostUtils.js
z3nL: Nope, nothing stopping us from using a Set - I thought about it but wasn't familiar with the syntax and went with the array in the interest of time

Thanks for bringing it up though - I think refactoring it into a Set would make the code look cleaner so that's def an easy to-do I can handle

client/src/utils/PostUtils.js
z3nL: Itu brought up this endpoint chaining in #14 ([comment](https://github.com/Zen-L-MU2025/project-sk8rlog/pull/15#discussion_r2193466565)) but the code you see here already existed before I read what he had to say about it

The first succeeding and the other failing steps into undefined behavior that I could mitigate with forced returns for now until I refactor all of these double endpoint call instances. 

Not sure if I interpreted your second question correctly but running these concurrently would just mean handling the logic in a single endpoint (which I think I'd prefer at this point), so nonetheless all signs point to "I should clean this up" 👍 thanks for the insight

client/src/utils/tokenization.js
z3nL: Heard you - one of those "why did I not think of that" moments lol

[[misc] fix regex var name](https://github.com/Zen-L-MU2025/project-sk8rlog/commit/6e5cb4fac7d591c629860b7d02c7e671a4b92b1e)

client/src/main/SinglePost.jsx
```diff
@@ -33,7 +55,14 @@ const SinglePost = () => {
             { post?.type === BLOGS && <img src={post?.fileURL} className='singlePostMedia' /> }
             <p>📍 {post?.location}</p>
             <p>{post?.description}</p>
-            <p>{post?.likeCount} likes</p>
+            <p className='likes'>
+                {`${postLikeCount} like(s)` }
```
stephanieprussick, line 59: Would be cool if you can implement it to say "1 like" or "2 likes" depending on the count!

z3nL: literally why did I not do this

[number of like(s)](https://github.com/Zen-L-MU2025/project-sk8rlog/commit/6d9e95b3b8845cd8a5a6a9c369bb390fb519d88d)

PR #17, opened 2025-07-10
[TC1] Posts now receive scores according to designed scoring formula

z3nL:
### Preface
I renamed _tokenization.js_ to _recUtils.js_ to keep file naming conventions consistent and its contents are now treated as a completely new file, so bare in mind that most of recUtils would be things we already reviewed in #16 

### Overview
The Home page served as the pilot for Task 2. This feature will also carry over to Posts, where users can sort by multiple filters that include "recommended"

When the home page loads and fetches posts to display, the scoring function is triggered
- Posts are filtered and scored based in the formula described in the TC1 documentation
- Posts are sorted by assigned scores in descending order
- If two posts have the same score, their comparison is resolved by creation date
- The resulting ordering of posts is delivered to the client using the corresponding state setter in the home page

### Test Plan
Console log of finalized post arrays with score attached to them before this update is sent to the client:
image


client/src/main/HomePage/HomePostsView.jsx
```diff
@@ -1,22 +1,27 @@
...
+import { scorePosts } from '/src/utils/recUtils'
 
 import '/src/css/home.css'
 
 const HomePostsView = ({ postType }) => {
-    const postType_lowercase = postType.toLowerCase()
+    const { activeUser } = useContext(UserContext)
 
-    const [posts, setPosts] = useState([])
+    const postType_lowercase = postType.toLowerCase()
```
stephanieprussick, line 16: nit - good to stick to the same format
you can use all camel case like postTypeLowerCase

client/src/utils/recUtils.js
```diff
@@ -0,0 +1,116 @@
...
+
+import axios from 'axios'
+import { removeStopwords, eng } from 'stopword'
+import { LIKE, NON_ALPHANUMERIC_REGEX, MILLISECONDS_IN_DAY, AGE_CUTOFF_IN_DAYS } from './constants.js'
+
+// Tokenize the content of a post, remove stop words
+// If the user has liked the post, increment the frequency of the tokens in the post; decrement if user is unliking
+// Store updated frequency in user.user_Frequency
+export const tokenize = async (post, activeUser, action) => {
+    const user_Frequency = activeUser.user_Frequency || {}
```
stephanieprussick, line 11: same comment, why not userFrequency?

client/src/utils/recUtils.js
```diff
@@ -0,0 +1,116 @@
...
+        for (const [tokenName, token] of Object.entries(overlap)) {
+            const base = token.frequency * userFrequency[tokenName].likedPostsPresentIn
+
+            // This token holds no weight, skip it
+            if (base === 0) {
+                continue
+            }
+
+            // Account for repetition factor and time factor
+            const repetitionFactor = 1 / ( userFrequency[tokenName].totalFrequencyAcrossLikedPosts / userFrequency[tokenName].likedPostsPresentIn )
```
stephanieprussick, line 86: is it possible that any of these could be 0?
 userFrequency[tokenName].totalFrequencyAcrossLikedPosts / userFrequency[tokenName].likedPostsPresentIn 
if so you would be doing 1/0 which surely will crash

client/src/utils/recUtils.js
```diff
@@ -0,0 +1,116 @@
...
+
+    // Sort posts by score; if two posts share the same score, sort by creation date
+    posts = posts.toSorted((a, b) => {
+        if (a.score === b.score) {
+            return (new Date(b.creationDate) - new Date(a.creationDate))
+        }
+        return b.score - a.score
+    })
+
+    console.log(posts)
```
stephanieprussick, line 106: nit - delete

z3nL: Only kept it in the PR because the test plan _is_ the console, unfortunately - removed as of #18

client/src/utils/recUtils.js
z3nL: Wow thanks for the catch here, in an old version of it 0 wasn't possible but now it is if a user unlikes the only post that had the relevant token(s)

[isNaN check in repetitionFactor](https://github.com/Zen-L-MU2025/project-sk8rlog/commit/426e873c8e8a38c36928ec2fd43d479908d54579)

client/src/utils/recUtils.js
z3nL: the object as a whole feels so unique from the rest of the user data that I wanted the naming to reflect that; also the column title in DB has the underscore so I didn't want to deviate

client/src/main/HomePage/HomePostsView.jsx
z3nL: Heard you - I added the underscore to indicate it as a "derivation" of postType rather than its own thing, but that type of nuance isn't something you can tell just by looking I imagine

z3nL: [rename postType_lowercase](https://github.com/Zen-L-MU2025/project-sk8rlog/pull/19/commits/074769393d665aa8452e0aaca03784edb1ed4038)

PR #18, opened 2025-07-10
[TC1][Viewing Recommendations] Recommendations visible clientside

z3nL:
### Overview
Following the implementation of scoring posts...
- the Home page now displays recommended posts on sign-in (see #17)
- the Clips and Blogs pages default to recommended posts
once filtering is added, users can swap between these recommended posts and other unbiased categories
- recommending posts verifiably do _not_ have an effect on posts visible on the Profile page, as a profile should just contain the full breadth of a user's posts

### Test Plan
[Video demonstration](https://www.loom.com/share/373702f675d341f3a98b1657eff8231c?sid=50019ae7-a608-4440-ac64-4f95156c4eba) of recommended content being displayed on sign-in. The Clips and Blogs pages visibly show this same content by default, while the Profile page includes _all_ posts from the user (which happens to be every post in the app, at the moment)

### DIFF
--- a/client/src/utils/PostUtils.js
+++ b/client/src/utils/PostUtils.js
@@ -3,6 +3,7 @@ const baseUrl = import.meta.env.VITE_API_URL || 'http://localhost:3000'
 import axios from 'axios'
 import { tokenize } from './recUtils.js'
 import { LIKE, UNLIKE } from './constants.js'
+import { scorePosts } from './recUtils.js'
 
 // Uploads a post, starting with the file attachment to GCS and then the full post data to server
 // Updates user's posts array state when complete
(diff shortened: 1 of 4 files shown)


stephanieprussick (approved):
looks good to me!

client/src/utils/recUtils.js
itumejia, line 1: not related to this PR, but I just noticed this. Let's improve naming for this file as it's hard to understand what you refer to with `rec`

client/src/utils/PostUtils.js
```diff
@@ -40,10 +41,15 @@ export const getUserPostsByType = async ( activeUser, postType, setUserPosts ) =
 }
 
 // Gets all posts by specified postType and sets the posts array state
-export const getAllPostsByType = async ( postType, setPosts ) => {
+// If scoringPayload is provided, will score the posts for the provided user and set the posts array state
+export const getAllPostsByType = async ( postType, setPosts, scoringPayload = { isScoring: false, activeUser: null } ) => {
```
itumejia, line 45: qq - in which cases we won't be scoring the posts?

z3nL: If the user wanted to sort the posts by a non-biased factor such as recency

Should I just always score the posts to minimize additional work needed when user changes the filter?

itumejia: ahh no, that makes sense! was just curious

PR #19, opened 2025-07-11
[Interaction] Displaying comments on a post and allowing users to add their own

z3nL:
### Overview
Pursuant to generating a popularity score for a post based on total likes and comments, the latter has been fully implemented as an interaction feature in the app
- Comments are displayed in ascending order in a scrollable box that starts at the bottom so users can see the latest comment
- Submitting a comment instantly populates it in the display
- If a post has no comments, the display indicates as such

### Test Plan
[Full walkthrough](https://www.loom.com/share/90dfc69f95a14955958f831ca4b0809c?sid=d959f76c-c8af-4086-b7c0-76477b8618bb) of:
- opening a post with no comments and adding your own
- opening a post with comments present and adding your own


client/src/main/SinglePost.jsx
```diff
@@ -16,43 +16,103 @@ import fullheart from '/src/assets/heartFull.png';
...
     const [post, setPost] = useState(null)
     const [postLikeCount, setPostLikeCount] = useState(0)
     const [postAuthor, setPostAuthor] = useState(null)
+    const [comments, setComments] = useState([])
+    const [commentBoxContent, setCommentBoxContent] = useState('')
+    const navigate = useNavigate()
+
+    const loadUser = async () => { await loadUserSession(setActiveUser) }
+
+    const HEADER_TEXT = 'Sk8rlog'
```
chris4307, line 30: can move to constants file

client/src/main/SinglePost.jsx
```diff
@@ -16,43 +16,103 @@ import fullheart from '/src/assets/heartFull.png';
...
         getPostByID(postID, setPost)
     }, [])
+
+    useEffect( () => {
+        verifyAccess(setHasAccess)
+    }, [activeUser])
+
+    useEffect( () => {
+        hasAccess === false && navigate('/unauthorized')
+    }, [hasAccess])
```
chris4307, line 45: could this be in the same useEffect as verifyAccess? it seems like if verifyAccess sets access to false then you just navigate to unauthorized.  Might be cleaner to put in the same effect

client/src/main/SinglePost.jsx
```diff
@@ -16,43 +16,103 @@ import fullheart from '/src/assets/heartFull.png';
...
+            <div className='singlePostComments'>
+                <h3>Comments</h3>
+                <div className='commentsBox'>
+                    { comments?.length === 0 && <p>None, yet!</p> }
+
+                    {  comments?.length > 0 &&
+                        comments?.map(comment => {
+                            return (
+                                <article className='comment' key={comment.commentID}>
+                                        <p className='commentDate'>{new Date(comment.creationDate).toLocaleDateString()}</p>
```
chris4307, line 98: nit: could we add a timestamp too if it's easy? if not then no need

client/src/main/SinglePost.jsx
```diff
@@ -16,43 +16,103 @@ import fullheart from '/src/assets/heartFull.png';
...
+                                        <p className='commentContent'><em>@{comment.author.username}</em>: {comment.content}</p>
```
chris4307, line 99: also could be cool to be able to like/reply to comments! This might be a pretty big feature though so focus on your other TCs first.

server/routes/posts.js
```diff
@@ -186,4 +186,77 @@ router.put('/:postID/likes/:action', async (req, res, next) => {
     }
 })
 
+// POST /posts/comments
+// Retrieves all comments corresponding to the comment IDs provided
+router.post('/comments', async (req, res, next) => {
```
chris4307, line 191: should comments be in its own file? something to think about

chris4307, line 191:
```suggestion
router.post('/comments', async (req, res, _next) => {
```

unused vars here can be preceded by a _ to show that they're unused

server/routes/posts.js
```diff
@@ -186,4 +186,77 @@ router.put('/:postID/likes/:action', async (req, res, next) => {
...
+        return res.status(STATUS_CODES.OK).json({ comments, message: 'Comments retrieved' })
+
+    } catch (error) {
+        return res.status(STATUS_CODES.SERVER_ERROR).json({ message: error })
+    }
+})
+
+// POST /posts/comments/:postID
+// Creates a comment for a post, saves it to the database and links it to its corresponding post
+router.post('/comments/:postID', async (req, res, next) => {
```
chris4307, line 227:
```suggestion
router.post('/comments/:postID', async (req, res, _next) => {
```

chris4307: same here

server/routes/posts.js
```diff
@@ -186,4 +186,77 @@ router.put('/:postID/likes/:action', async (req, res, next) => {
...
 })
 
+// POST /posts/comments
+// Retrieves all comments corresponding to the comment IDs provided
+router.post('/comments', async (req, res, next) => {
+    try {
+        const { commentIDs } = req.body
+
+        if (commentIDs?.length < 1) {
+            return res.status(STATUS_CODES.BAD_REQUEST).json({ message: 'No comment IDs provided' })
```
chris4307, line 196: is this necessarily a bad request if we just get no commentIDs? For posts that actually don't have comments this would be expected behavior

server/routes/posts.js
```diff
@@ -186,4 +186,77 @@ router.put('/:postID/likes/:action', async (req, res, next) => {
...
+        for ( const commentID of commentIDs ) {
+            const comment = await prisma.comment.findUnique({
+                where : { commentID }
+            })
+
+            // Retrieve the author of the comment and attach it to the comment object
+            const author = await prisma.user.findUnique({
+                where : { userID : comment.authorID }
+            })
+            comment["author"] = author
```
chris4307, line 210: this actually makes sense given a user might change their username (over storing their username in the comments object).

However, there are two ways I could see you reducing the number of DB requests

1. cache the author usernames so if you get a hit you dont have to hit prisma again
2. batch the request (if possible).

server/routes/posts.js
```diff
@@ -186,4 +186,77 @@ router.put('/:postID/likes/:action', async (req, res, next) => {
...
+        })
+
+        // Link the comment to the post through its ID
+        const post = await prisma.post.findUnique({
+            where: { postID }
+        })
+        await prisma.post.update({
+            where: { postID },
+            data: { comments: [...post.comments, comment.commentID] }
+        })
```
chris4307, line 246: quick idea, since you're doing this async (and you render the comment apart from this anyway), does it make sense to sort the comments list before committing to the DB, so you don't have to sort every time you load the comments?

client/src/utils/PostUtils.js
```diff
@@ -145,11 +145,37 @@ const unlikePost = async ( post, activeUser, setActiveUser ) => {
...
+            console.error("getComments error: ", error)
+        })
+}
+
+// Creates a comment in DB and adds it to the comments array state
+export const createComment = async (commentBoxContent, activeUser, postID, setComments) => {
+    const userID = activeUser.userID
+    await axios.post(`${baseUrl}/posts/comments/${postID}`, { commentBoxContent, userID })
+        .then(res => {
+            setComments(comments => [res.data.comment, ...comments])
```
chris4307, line 176: maybe you could optimistically set the comments to include the one you just submitted, and only if the DB put errors out you remove the comment?

chris4307 (commented):
Looks great! Just left a couple comments, mostly relating to perf considerations.

client/src/main/SinglePost.jsx
z3nL: HEADER_TEXT varies by page that uses header, should I just make a different Header Text constant for each and keep them all in the file

client/src/main/SinglePost.jsx
z3nL: I think it looks better that way, too

[condense useEffects](https://github.com/Zen-L-MU2025/project-sk8rlog/commit/3ff1766f76409291893c8c6c9b2e3c3ca3c4b0df)

client/src/main/SinglePost.jsx
z3nL: Sure! 

[timestamp on comment](https://github.com/Zen-L-MU2025/project-sk8rlog/commit/5b416753829bf567ad10bfc09ae7d12a9b523bc8)

client/src/main/SinglePost.jsx
z3nL: It's in the backlog but yeah figured it's more of a nice-to-have at the moment

server/routes/posts.js
z3nL: I considered it but the commenting is so basic that I treated it as an extension of posts; on the topic of liking/replying to comments I think that'll definitely be cause to move comments into its own file

z3nL: Wow I had no idea that was a thing, cool

[mark next as unused](https://github.com/Zen-L-MU2025/project-sk8rlog/commit/80a9ff2d86f3fb6f7642a667d0dbaab591ed9aae)

server/routes/posts.js
z3nL: The utility function that calls this endpoint is prefaced with
`if (!commentIDs?.length) return`
to frame it as "don't waste time calling the endpoint if there aren't any comments to fetch"
So if the endpoint is reached without any comments to look up then that would be an issue, if that makes sense

maybe I should just isolate the query handling in the backend exclusively to improve visibility on this?

server/routes/posts.js
z3nL: Yeah that was my intent but good point about caching the author names; what is batching a request? I'm unfamiliar with the term 🙃

server/routes/posts.js
z3nL: But If I sort the comments list before committing to the DB isn't that the same as sorting it when I load them? Or do you mean sort the comments in DB once and refactor so sorting won't _ever_ have to happen again (i.e. just append to the end of the existing comments list)

my bad if I'm misunderstanding

client/src/utils/PostUtils.js
z3nL: Or maybe if creating the comment fails I could throw a screen alert in the catch to indicate as such?

client/src/main/SinglePost.jsx
chris4307: ah hm, ya up to you then if it's always different then you can just keep it as is

client/src/main/SinglePost.jsx
chris4307: sweet!

client/src/main/SinglePost.jsx
chris4307: definitely. Don't worry about this if you have other hi prio stuff

server/routes/posts.js
chris4307: it's just a naming convention! doesn't really actually do anything haha. Some linters will have you add the underscore.

server/routes/posts.js
chris4307: oh ok sure, ya i mean if you avoid even calling it on the client if there are no commentIDs then that should be ok. Just remember every time you use this endpoint you'll need that check in the client, so next time you use this endpoint if it makes sense you can just move the check to the backend

server/routes/posts.js
chris4307: just including a list of commentIDs in a single request payload. So if you have 10 commentIDs, making 1 network request instead of 10, for example.

server/routes/posts.js
chris4307: no worries! functionally it's the same, but if you think about where the critical load path is, on load you need to wait for the comments to load, and you want it to be as fast as possible.

on comment submit, the timing of submitting the comment is not as critical.

Often _when_ you do the computation work matters for load performance.

For the number of comments that are being loaded, this probably is never going to really be an issue, but just an interesting exercise for you :)

You can also append to the comments list, but beware of race conditions if 2 people submit at the same time. But it is much faster than sorting every time.

client/src/utils/PostUtils.js
chris4307: true! most websites will do something like update the comment on the UI indicating that it didn't upload as a non-blocking error with a retry button. This sounds out of scope so whatever makes sense to you works

server/routes/posts.js
z3nL: When you put it like that it makes a lot more sense, thanks a bunch

PR #22, opened 2025-07-14
[TC1][Scoring] Video Length as a Factor in Post Length Bias

z3nL:
### Overview
The distinction between Clips and Blogs in my app made it seem disingenuous to only account for description lengths when Clips have full videos to view in them.
- When calculating a clip's post length, we pass its fileURL to [a library](https://www.npmjs.com/package/get-video-duration) that calculates its length in seconds
- Using this length, we translate it into a "word count," in the same way we find the word count of the post's description, by multiplying it by the generally agreed on average reading speed in words per minute, converted into words per second. This length is added to the post's overall length in calculations

#### Caveat
Calculating the length of every video each time the feed loads again results in a lot of overhead that slows down the app - a future PR will have updated the backend structure so that the length of a clip is saved in the database on post creation

### Test Plan
**See yellow text on right side of screen representing data for first / second post cards on left
To confirm that a clip's calculated length differs from its description word count and subsequently affects its post length bias standing, I checked their stored post lengths in the component tree. As expected, the visible post length is much more than just its description.
image
image

### DIFF
--- a/client/src/utils/recommendationUtils.js
+++ b/client/src/utils/recommendationUtils.js
@@ -104,6 +104,7 @@ export const scorePosts = async (posts, activeUser, setPosts, isByPopularity) =>
         let postLength = new String(post.description).split(NON_ALPHANUMERIC_REGEX).length
         if (post.type === CLIPS) {
             const res = await axios.post(`${baseUrl}/posts/clipLength`, { fileURL : post.fileURL})
+                .catch(error => console.error(error))
             const clipLength = res.data.clipLength
             const clipLengthAsWordCount = Math.ceil(clipLength * AVERAGE_WORDS_READ_PER_SECOND)
             postLength += clipLengthAsWordCount
--- a/client/src/utils/recommendationUtils.js
+++ b/client/src/utils/recommendationUtils.js
@@ -157,7 +157,11 @@ const calculateBiasFactors = async (activeUser) => {
 
         if (post?.type === CLIPS) {
             likedClips++
-            // TODO Implement advanced video length logic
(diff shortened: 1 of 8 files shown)


client/src/utils/recommendationUtils.js
```diff
@@ -97,8 +99,20 @@ export const scorePosts = async (posts, activeUser, setPosts, isByPopularity) =>
             rawPostScore += tokenScore
         }
 
+        // Split the description into tokens representing full words as a post's length
+        // Standardize video length as "additional reading" in a post
+        let postLength = new String(post.description).split(NON_ALPHANUMERIC_REGEX).length
+        if (post.type === CLIPS) {
+            const res = await axios.post(`${baseUrl}/posts/clipLength`, { fileURL : post.fileURL})
```
z3nL, line 106: I plan on scrapping this in favor of clip length being stored on post creation and I used this exact block twice so just leaving it on the record that this is going to be pocketed into its own function

client/src/utils/recommendationUtils.js
```diff
@@ -145,7 +159,11 @@ const calculateBiasFactors = async (activeUser) => {
 
         if (post?.type === CLIPS) {
             likedClips++
-            // TODO Implement advanced video length logic
+            const res = await axios.post(`${baseUrl}/posts/clipLength`, { fileURL : post.fileURL})
+                .catch(error => console.error(error))
```
z3nL, line 163: The second instance

z3nL (commented):
General comments / req feedback if I come up with any

itumejia (commented):
LGTM, I think storing the video length is a good idea and would have great benefit in performance! Although I'd say it's not critical for now

PR #23, opened 2025-07-15
[UI][Bug] Fixed Post Loading Incorrectly on Creation

z3nL:
### Overview
Previously on the user's profile page, a post would be created successfully but fail to display its file upload in the preview card that populates the screen when the user submits. 

This has been resolved through enabling CORS in the app's Google Cloud Storage bucket on server startup so that the client continuously fetches the newly created post's file URL using the HEAD method until it confirms that the data within it is ready to access and only then displays the corresponding card.

### Additional Plans
This lays groundwork for a significant performance improvement in TC1. Waiting for the file URL to be ready will also be applied in the backend so that endpoints can store clip video lengths on creation instead of repeatedly computing them in the recommendation algorithm.

### Test Plan
[Sample video](https://www.loom.com/share/2b7beee954424e0081ce083a64f9ec18?sid=81707848-e12c-4f88-be92-85bd17c70836) of user uploading a clip/blog and having it successfully populate the screen if the corresponding type view is open

### DIFF
--- a/client/src/main/PostCard.jsx
+++ b/client/src/main/PostCard.jsx
@@ -1,7 +1,7 @@
-import { useRef, useContext } from 'react'
+import { useRef, useContext, useEffect, useState } from 'react'
 import { Link } from 'react-router'
 
-import { deletePost, handleLikeOrUnlikePost } from '/src/utils/postUtils'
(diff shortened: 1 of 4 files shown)


itumejia (commented):
nice finding! as discussed yesterday, it would be great if you could get pull some learnings and share them in a post

PR #21, opened 2025-07-14
[TC1][Scoring] Additional Factors in Recommendation Algorithm Accounted For and Reflected in Client

z3nL:
# Overview
### Recommendation Complexity
In addition to recommending posts based on frequency of tokenized content and post age, a post's final score now incorporates a bias factor that includes the following tangibles:
- Popularity, decided by a weighted total of likes and comments
- Type bias, the proportion of the user's liked posts that are the same type as the post being assessed
- Post length bias, the percent deviation of the post's content length from the average content length of the user's liked posts

### Sorting with new recommendation factors in client
Users can now sort displayed posts by
- Recommended (calculated overall score)
- Latest (date, descending)
- Popularity (calculated popularity score)
- Near You is present in the dropdown, but it is defaulted to the same functionality as Latest until locations are implemented in the app

# Test Plan
[Video Link](https://www.loom.com/share/0b9d9a125c6b485d8d62114373565a0c?sid=f30ba7d4-1fd1-411f-acc5-2945fc4f4411)
### Sample video of:
- Sorting Clips. The clip the user liked is recommended first, but the unliked one has many comments and takes precedence in popularity
- The same functionality in Blogs, but with more posts present to allow for variation. Note how checking Near You yields the same result as Latest

PR #24, opened 2025-07-15
[Gen] Code Cleanup To Address Feedback

z3nL:
### Overview
#19 and #21 were very content-heavy and plenty of feedback to consider and implement followed. This PR just aggregates these varied changes to keep them separate from my work on TC2 so far.


itumejia:
Although no actual functionality was added in this PR, it's always a good idea to add some preventive test plan to make sure no side effects are caused. For example, since you are mainly cleaning up code from the ranking algorithm logic, a quick screen recording showing that the main feed still loads correctly would be enough to make sure nothing is broken with this :)

PR #27, opened 2025-07-16
[TC2][Notifications] Dummy System Implemented

z3nL:
### Overview
Before profiles can be ranked to suggest to users in a notification, the notification system itself must be implemented. This PR implements the skeleton of the system with functionality to verify its performance.

- The app uses the socket.io library to establish a client-server WebSocket connection, which will facilitate the server delivering suggestion notifications to the client in the future.
- The connection is established on initial render of client (root component App.jsx) and logs a confirmation message to client console stating as such
- The server socket endpoint hosts a dummy notification delivery system to simulate real delivery. 
-- Every five seconds, it decides whether fire a notification by randomly selecting a number. 
-- If the number is above a defined threshold (representing percent chance of firing), the server emits a notification delivery event containing a message denoting the time of firing.
- The client socket endpoint listens for notification delivery events. When the event triggers, it stores the content in an array state defined in the root component

### Test Plan
- [Video demonstration](https://www.loom.com/share/c8dde5c12db0453cbeff3882308d579a?sid=0c33503d-de94-4b9d-ba56-c2306b7c17ea) of websocket connection being established (confirmed in console) and the _dummyNotifications_ array state autonomously populating as the server fires notification events.


client/src/utils/recommendationUtils.js
```diff
@@ -113,7 +112,7 @@ export const scorePosts = async (posts, activeUser, setPosts, scoringMode) => {
 
     // Sort posts by either recommendation score or popularity
     // Let tie breakers be handled by other option and finally by creation date
-    posts = sortByMetric(posts, scoringMode)
+    posts = await sortByMetric(posts, scoringMode)
```
itumejia: why did we add await here?

server/utils/notifications.js
itumejia, line 1: nit: let's name the file `notificationsUtils` to keep consistency? same for websockets

client/src/App.jsx
```diff
@@ -9,14 +9,19 @@ import Posts from './main/Posts'
...
 
 import UserContext from '/src/utils/UserContext.js'
+import { establishWebSocketConnection } from '/src/utils/websocket.js'
 import { CLIPS, BLOGS } from '/src/utils/constants'
 
 import './App.css'
 
 function App() {
-
+  const [dummyNotifications, setDummyNotifications] = useState([])
```
itumejia: although notifications are "dummy" now, I'm guessing we'll be using the same skeleton you're building here to keep adding functionalities on top? Then let's avoid any naming with 'dummy' in it to keep it clean from the start

client/src/utils/recommendationUtils.js
z3nL: I can't recall; in all honesty I've developed a response habit where if my asynchronous functions aren't working as intended then I blame forgetting an await somewhere 😭 

just verified it's redundant, which tracks since toSorted is built in and holds off until completion

[undo needless await](https://github.com/Zen-L-MU2025/project-sk8rlog/commit/74d10618b46316f9a773c8f4fcfef47d5d63259b)

server/utils/notifications.js
z3nL: Fair

[more naming consistency](https://github.com/Zen-L-MU2025/project-sk8rlog/pull/27/commits/f549592131fe861133e7f85666cddf3e18fa0734)

client/src/App.jsx
z3nL: Good point, I wanted this PR to hold that emphasis in naming too but I guess [it's not dumb, actually](https://github.com/Zen-L-MU2025/project-sk8rlog/pull/27/commits/bea997fcccfb451dc34bff27c2a3407eab604c00)

PR #29, opened 2025-07-18
[Gen] Semicolons

z3nL:
### Overview
My codebase did not use semicolons. I thought that semicolons would be seen as redundant because statements in JS are complete without them, but it was brought to my attention that it is better practice to include them.
- Used VS Code Prettier extension to quickly reformat each file in the codebase

### Test Plan
TBD

What would be a reasonable test plan for a change like this? The app expectedly works the same and I'm not sure if creating a video to prove that is needed here


chris4307:
For these types of changes, just showing that the app still works should be sufficient. with a prettier reformat, I would guess that for the most part everything should work or everything should break.

PR #34, opened 2025-07-19
[Gen][TC2] Following Users

z3nL:
### Overview
Pursuant to the mutual following metric of a follow suggestion candidate in TC2, users can now follow/unfollow each other on the app by going to a profile and hitting the corresponding button
- A users' following list is preserved in the database and dynamically updated on button press so that it persists across sessions

### Test Plan
[Sample video](https://github.com/user-attachments/assets/d1fcbb26-7ef2-4951-b9d0-328a0669014a) of a user following another, refreshing the page, and unfollowing

User following list in database for demo user (third column) updates accordingly after each action
image
image
image

### DIFF
--- a/client/src/main/ProfilePage/ProfileHead.jsx
+++ b/client/src/main/ProfilePage/ProfileHead.jsx
@@ -15,6 +15,7 @@ const ProfileHead = ({ userToDisplay, activeUser, setProfileContentView, toggleC
                 </p>
                 <p>📍 {userToDisplay?.location || "Location not set"}</p>
                 <p>Bio: {userToDisplay?.bio || "No bio provided"}</p>
+                {!isSelfProfile && <button className="followButton">+ Follow User</button>}
             </div>
             <div className="contentButtons">
                 <p className="contentButton" onClick={() => setProfileContentView(CLIPS)}>
--- a/client/src/main/ProfilePage/ProfileHead.jsx
+++ b/client/src/main/ProfilePage/ProfileHead.jsx
@@ -1,10 +1,20 @@
-import { CLIPS, BLOGS } from "/src/utils/constants";
+import { useContext } from "react";
+
+import UserContext from "/src/utils/UserContext";
+import { handleUserFollowing } from "/src/utils/userUtils";
+import { CLIPS, BLOGS, FOLLOW, UNFOLLOW } from "/src/utils/constants";
+
 import skateboard from "/src/assets/skateboard.png";
 
 import "/src/css/profile.css";
 
-const ProfileHead = ({ userToDisplay, activeUser, setProfileContentView, toggleCreatePostModal }) => {
+const ProfileHead = ({ userToDisplay, setProfileContentView, toggleCreatePostModal }) => {
+    const { activeUser, setActiveUser } = useContext(UserContext);
+
     const isSelfProfile = userToDisplay?.userID === activeUser?.userID;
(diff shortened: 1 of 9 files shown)


client/src/main/ProfilePage/ProfileHead.jsx
itumejia, line 1: not necessary to do this, but a piece of advice when something your building keeps scaling up, it's a good idea to extract components along with their logic to keep your codebase polished. For example, this 'follow button' you are adding, could be extracted into a different component, that way you'd also keep the 'isFollowing' logic isolated from the main ProfileHead component

itumejia (commented):
LGTM!

client/src/main/ProfilePage/ProfileHead.jsx
z3nL: Sounds good, I'll try this out some time, thanks 🙏

client/src/utils/userUtils.js
```diff
@@ -191,3 +187,25 @@ const setCookies = async (token, userID, setIsSuccessful = null) => {
...
+export const handleUserFollowing = async (activeUser, userBeingReferencedID, action, setActiveUser) => {
+    await axios
+        .put(`${baseUrl}/users/${activeUser.userID}/followedUsers/${action}`, { userBeingReferencedID })
+        .then(() => {
+            const updatedUser = {
+                ...activeUser,
+                followedUsers:
+                    action === FOLLOW
+                        ? [...activeUser.followedUsers, userBeingReferencedID]
+                        : activeUser.followedUsers.filter((uID) => uID !== userBeingReferencedID),
```
chris4307, line 201: as this is just a string array, you could also consider using a set here

server/routes/users.js
```diff
@@ -132,4 +139,28 @@ router.put("/:userID/likedPosts/:action", async (req, res, _next) => {
...
+
+        const user = await prisma.user.findUnique({ where: { userID } });
+        await prisma.user.update({
+            where: { userID },
+            data: {
+                followedUsers:
+                    action === FOLLOW
+                        ? [...user.followedUsers, userBeingReferencedID]
+                        : user.followedUsers.filter((uID) => uID !== userBeingReferencedID),
+            },
```
chris4307, line 158: seems like there's some duplicate code here as in userUtils, any way to extract this into a util function? Similar idea to what Ituriel said, but as projects scale up, duplicate definitions can become a huge pain down the line (obviously won't be a problem for this project, but something that's great to keep in mind when you start working on large projects :))

chris4307 (commented):
LGTM, looks great!

client/src/utils/userUtils.js
```diff
@@ -191,3 +187,25 @@ const setCookies = async (token, userID, setIsSuccessful = null) => {
             return;
         });
 };
+
+// Adds/removes a user from active user's following list
+export const handleUserFollowing = async (activeUser, userBeingReferencedID, action, setActiveUser) => {
+    await axios
+        .put(`${baseUrl}/users/${activeUser.userID}/followedUsers/${action}`, { userBeingReferencedID })
+        .then(() => {
+            const updatedUser = {
```
chris4307, line 196: was actually looking to see how you handle a DB write failure. Great thinking on conditionally updating the state! The alternative would be optimistically update the state before the write, but that adds complexity in needing to undo it on DB write failure.

PR #35, opened 2025-07-21
[TC2] Profile Ranking Algorithm Complete

z3nL:
### Overview
In #33, I created and endpoint to test a default output of the profile ranking algorithm. Pursuant to the outline described in [documentation](https://docs.google.com/document/d/1kNEz32Xm4RZwRLB867fBea2WWoBEhev3rD0eiksIfTs/edit?tab=t.0#heading=h.plj4hngf5wup), I implemented comparing users' interests and cross referencing who they follow (#34), henceforth completing the algorithm.

The nature of NLP computation that needs to be performed to compare user interests ushered in the need for some work in Python - it could've all been done in Node, but it's much less common and subsequently makes obtaining concrete documentation and implementation harder than it needs to be. Check out [my findings on this](https://docs.google.com/document/d/1kNEz32Xm4RZwRLB867fBea2WWoBEhev3rD0eiksIfTs/edit?tab=t.0#heading=h.igvtuwiutuj3) for more context on this decision and what's going on in _app.py_

- created a Flask application in Python that serves the sole purpose of generating user frequency object embeddings and calculating cosine similarity scores
- **evaluateCandidate** in _profileRankingUtils_ invokes the Flask endpoint _/embedUserFrequency_, passing in host and candidate keyword arrays and the user_Frequency objects they originated from to generate repetition factor weights
- Using the _gensim_ library, the script loads the Google News Dataset Word2Vec model
- Each keyword in a user input array gets mapped to a vector that subsequently has the repetition factor weight applied to it
- After vectorizing each keyword, the array of vectors is averaged out into a single embedding that "describes" the user's interests
- The host and candidate embeddings are compared using cosine similarity with _numpy_ linear algebra utilities
- The interest embeddings and cosine similarity score are sent back as a response to **evaluateCandidate**, which uses these computations to finalize calculations alongside mutual following frequency
- The _/acquireCandidates_ endpoint subsequently returns data with the same behavior as originally implemented - an array of all candidates in descending order of _candidacyScore_

### Test Plan
Another [test run](https://github.com/user-attachments/assets/13f4f8ce-ba22-40b6-b069-52bbc44883e5) of the _/acquireCandidates_ endpoint, where visibly different candidacy scores are being output now that the whole algorithm has come together. Outside of endpoint functionality, all metrics have been and are being continuously console log tested to ensure functionality as calculations are performed.

### DIFF
--- a/server/app.py
+++ b/server/app.py
@@ -1,0 +1,72 @@
+from flask import Flask, request, jsonify
+from gensim.models import KeyedVectors
+from numpy import dot
+from numpy.linalg import norm
+import numpy as np
+
+# Load the Google News word2vec model
+model_path = 'GoogleNews-vectors-negative300.bin'
+model = KeyedVectors.load_word2vec_format(model_path, binary=True)
+
+# Vectorizes each keyword in a user frequency object and averages out all vectors as an embedding for that particualr user
+def vectorize(keywords, frequencies):
+    vectors = []
+    collectiveWeight = 0
+
+    for i, word in enumerate(keywords):
+        if word in model:
+            vector = model[word]
+
+            # Compute weight based on repetition factor and add to collective weight
+            repetitionWeight = 1
+            if frequencies[word]["likedPostsPresentIn"] != 0:
+                repetitionWeight += 1 / (frequencies[word]["totalFrequencyAcrossLikedPosts"] / frequencies[word]["likedPostsPresentIn"])
+            collectiveWeight += repetitionWeight
+
+            # Add vector to list with weight applied
+            vectors.append(vector * repetitionWeight)
+
+        # Otherwise, the word just isn't in the model, nothing to do here (yet?)
+
+    if not vectors or collectiveWeight == 0:
+        return np.zeros(model.vector_size)
+
+    # Average out all vectors as one
+    averageOfVectors = np.sum(vectors, axis=0) / collectiveWeight
+    return averageOfVectors
+
+
+app = Flask(__name__)
+
+# Default
+@app.route("/")
+def root():
+    return "ROOT"
+
+# Create vector embeddings for user and candidate frequency objects and send results in response
+@app.route("/embedUserFrequency", methods=['POST'])
+def embedUserFrequency():
+    data = request.get_json()
+    # Vectorization returns a NumPy array, standardize this to a normal array that can be processed into JSON using tolist()
+    userFrequencyEmbedding = vectorize(data["userKeywords"], data["userFrequency"]).tolist()
+    candidateFrequencyEmbedding = vectorize(data["candidateKeywords"], data["candidateFrequency"]).tolist()
+
+    # Compute cosine similarity score as dot product of embeddings divided by the product of their normal vectorizations
+    userNorm = norm(userFrequencyEmbedding)
+    candidateNorm = norm(candidateFrequencyEmbedding)
+
+    # If either user has never liked a post, fall back to 0
+    if userNorm == 0 or candidateNorm == 0:
+        cosineSimilarityScore = 0
+    else:
+        cosineSimilarityScore = dot(userFrequencyEmbedding, candidateFrequencyEmbedding) / (userNorm * candidateNorm)
+
+    return jsonify({
+        "userFrequencyEmbedding" : userFrequencyEmbedding,
+        "candidateFrequencyEmbedding" : candidateFrequencyEmbedding,
+        "dimension" : len(userFrequencyEmbedding),
+        "cosineSimilarityScore" : cosineSimilarityScore
+    })
+
+if __name__ == "__main__":
+    app.run()
--- a/server/app.py
+++ b/server/app.py
@@ -46,7 +46,9 @@ def root():
 # Create vector embeddings for user and candidate frequency objects and send results in response
 @app.route("/embedUserFrequency", methods=['POST'])
 def embedUserFrequency():
+    # Node translation: data = req.body
     data = request.get_json()
+
     # Vectorization returns a NumPy array, standardize this to a normal array that can be processed into JSON using tolist()
     userFrequencyEmbedding = vectorize(data["userKeywords"], data["userFrequency"]).tolist()
     candidateFrequencyEmbedding = vectorize(data["candidateKeywords"], data["candidateFrequency"]).tolist()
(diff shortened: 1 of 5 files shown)


server/app.py
itumejia, line 1: It's fine since you probably won't iterate much in this Python component, but it would keep it clean if you modularize code here a bit so that not everything is thrown in a single file. Ideally having at least one init `app` file, and another one with the actual endpoints

server/utils/profileRankingUtils.js
```diff
@@ -52,17 +56,37 @@ const evaluateCandidate = async (user, candidate) => {
...
+        method: "POST",
+        headers: {
+            "Content-Type": "application/json",
+        },
+    })
+        .then((res) => res.json())
+        .then((data) => {
+            userEmbedding = data.userEmbedding;
+            candidateEmbedding = data.candidateEmbedding;
+            cosineSimilarityScore = isNaN(data.cosineSimilarityScore) ? 0.1738 : data.cosineSimilarityScore;
```
itumejia, line 82: `0.1738` let's pull out this number into a constant?

server/utils/profileRankingUtils.js
```diff
@@ -52,17 +56,37 @@ const evaluateCandidate = async (user, candidate) => {
         popularityOvr = totalPopularity / candidatePosts.length;
     }
```
itumejia, line 58: it's getting a bit hard to read this function. It'd help if you modularized code a bit by extracting some of the different steps you are taking here into functions. 

For example, you could have a function called `getSimilarityFactor` that takes `user` and `candidate` as params, does all the fetching, and returns the similarityFactor. You could do a similar thing for other factors you are using

z3nL: Admittedly I got a bit carried away there
image

Pushed and verified to function the same: [[misc] Modularize profile ranking utils](https://github.com/Zen-L-MU2025/project-sk8rlog/pull/36/commits/7ed3c721ea3419122ed00f23df1551fb15ea42c4)

server/utils/profileRankingUtils.js
z3nL: Residual magic number from testing that I missed, my bad - see [[misc] Modularize profile ranking utils](https://github.com/Zen-L-MU2025/project-sk8rlog/pull/36/commits/7ed3c721ea3419122ed00f23df1551fb15ea42c4)

server/app.py
z3nL: Heard you - I didn't for the same reason you inferred but I should still be working in the codebase's best interest, if I have to iterate on the Python component I'll def modularize it

server/utils/profileRankingUtils.js
```diff
@@ -71,9 +95,10 @@ const evaluateCandidate = async (user, candidate) => {
 
     // Hold this overlap against all users they follow to create a mutual following factor
     const followingUnion = new Set(userFollowing.concat(candidateFollowing));
-    const mutualFollowFrequency = followingOverlap.length / followingUnion.size;
+    let mutualFollowFrequency = followingOverlap.length / followingUnion.size;
+    mutualFollowFrequency = isNaN(mutualFollowFrequency) ? 0 : mutualFollowFrequency;
```
chris4307, line 99: I'm assuming the isNaN check is for divide by 0?

Might be easier to read like this. Might not be easier. I'll let you decide haha. Also depends if you want to check for null and 0

```
const mutualFollowFrequency = followingUnion.size===0 ? 0 : followingOverlap.length / followingUnion.size;
```

server/utils/profileRankingUtils.js
```diff
@@ -52,17 +56,37 @@ const evaluateCandidate = async (user, candidate) => {
...
+            candidateEmbedding = data.candidateEmbedding;
+            cosineSimilarityScore = isNaN(data.cosineSimilarityScore) ? 0.1738 : data.cosineSimilarityScore;
+        })
+        .catch((err) => {
+            console.log(err);
+        });
 
     // Convert the similarity score to a scalar factor
-    const similarityFactor = similarityScore < 0 ? -1 + similarityScore : 1 + similarityScore;
+    const similarityFactor = cosineSimilarityScore < 0 ? -1 + cosineSimilarityScore : 1 + cosineSimilarityScore;
```
chris4307, line 89: nit: there might be a cleaner way to do this since it's basically (1 + abs(cosineSimilarityScore))*(-1 or 1)

chris4307: also, what happens if the fetch earlier fails? cosineSimilarityScore won't be a number.

server/utils/profileRankingUtils.js
```diff
@@ -52,17 +56,37 @@ const evaluateCandidate = async (user, candidate) => {
...
-    // then vectorize
-    // then pass insto cosine similarity function
-    const similarityScore = 0.1738; // placeholder
+    const userFrequency = user.user_Frequency || {};
+    const candidateFrequency = candidate.user_Frequency || {};
+
+    const userKeywords = Object.keys(userFrequency);
+    const candidateKeywords = Object.keys(candidateFrequency);
+
+    let userEmbedding, candidateEmbedding, cosineSimilarityScore;
```
chris4307, line 68: do userEmbedding and candidateEmbedding need to be defined by a `let`?

server/utils/profileRankingUtils.js
```diff
@@ -1,12 +1,14 @@
...
+import { RANKING_MODES, POST_OVR_WEIGHT, POPULARITY_OVR_WEIGHT } from "./constants.js";
 import { scorePosts } from "./serverPostRecommendationUtils.js";
 
 const prisma = new PrismaClient();
 
 // Rank suggestion candidates for some user and return candidates in descending order of score
 export const rankCandidates = async (hostUserID) => {
     const user = await prisma.user.findUnique({ where: { userID: hostUserID } });
+    // No passwords
+    delete user.password;
```
chris4307, line 11: Out of scope for this project, but passwords should probably be stored in a different table than the rest of the user data. Not sure if you did this already but they should also be encrypted at rest and in transit (honestly all user data probably should but again, out of scope here)

server/utils/profileRankingUtils.js
z3nL: Yeah, it is for divide by 0 but what you wrote is pretty easy to read since null shouldn't be a result of operations here

server/utils/profileRankingUtils.js
z3nL: Agreed and great catch thanks, I defaulted cosineSimilarityScore to 0 in the catch to mitigate

server/utils/profileRankingUtils.js
z3nL: Nah, it just made sense in my head to write them all as _let_ to indicate "this is what we're gonna find", should I generally avoid writing lets if the variable is only assigned a value once?

server/utils/profileRankingUtils.js
z3nL: Slowly realizing exactly that haha, the separate table might've been better; passwords are encrypted on account creation using bcrypt but I just omit the field entirely when working with a user object out of intuition

PR #37, opened 2025-07-21
[TC2] Notifications Scheduling - Reached Working State

z3nL:
### Overview
The cron-based notification scheduling system has reached working functionality pursuant to some of the objectives outlined in [documentation](https://docs.google.com/document/d/1kNEz32Xm4RZwRLB867fBea2WWoBEhev3rD0eiksIfTs/edit?tab=t.0#heading=h.5z5nkj7lixr7)

At regular intervals during the server lifespan, we now:
- Sweep the database for all users whose best notification times align with the time the sweep was conducted
-- Leverages a half-hour window of allowance both in the past and present to count in alignment

- For each "notifiable" user, perform a candidate search
- Emit a notification through the WebSocket confirming that candidate Y was successfully suggested to user X

### Completion Plans
- Implement a _recentlySuggestedUsers_ column in the User model to track if it would be redundant to suggest some candidate again because it already happened too recently
- Implement socket rooms for each user signed into the app such that the cron job only fires notifications to the room the suggestion notification actually belongs to

### Current Test Plan
Sample user _@/x_ was deliberately created so that their session data would align with the cron jobs firing at the general time of account creation. 
- Notifications would successfully fire a suggestion to follow _@/z3nL_ while the lower half-hour bound checked out
- However, as the current time exit the lower window bound, notifications expectedly stopped firing, serving as proof of proper time checks and subsequent delivery

Notification logs, in descending order of recency top to bottom
image
image

### DIFF
--- a/client/src/utils/clientWebSocketUtils.js
+++ b/client/src/utils/clientWebSocketUtils.js
@@ -8,9 +8,7 @@ export const establishWebSocketConnection = (setNotifications, setHasNewNotifica
     const identifier = new Date().getMilliseconds();
 
     const socket = io(VITE_SERVER_URL);
-    socket.on(CONNECT, () => {
-        console.log(`Websocket connection established at ${new Date().toLocaleTimeString()}`);
-    });
+    socket.on(CONNECT, () => {});
 
     socket.on(DISCONNECT, () => {
         console.log(`disconnected at ${new Date().toLocaleTimeString()}`);
--- a/server/utils/notificationSchedulingUtils.js
+++ b/server/utils/notificationSchedulingUtils.js
@@ -1,19 +1,83 @@
 import cron from "node-cron";
-import { CONNECTION, DISCONNECT, DELIVER_NOTIFICATION, CRON_INTERVAL_STRING, CRON_INTERVAL_DESCRIPTOR } from "./constants.js";
+import { toSecondOfDay } from "./sessionUtils.js";
+import {
+    CONNECTION,
+    DISCONNECT,
+    DELIVER_NOTIFICATION,
+    CRON_INTERVAL_STRING,
+    CRON_INTERVAL_DESCRIPTOR,
+    HALF_HOUR_IN_SECONDS,
+    MIDDLE_OF_DAY_IN_SECONDS,
+} from "./constants.js";
+import { PrismaClient } from "../generated/prisma/index.js";
+const prisma = new PrismaClient();
 
 export const requestNotificationScheduling = (socketServer) => {
     console.log(`Cron generated - running a task every ${CRON_INTERVAL_DESCRIPTOR}`);
 
     socketServer.on(CONNECTION, (socket) => {
-        console.log("notiSched utils: connection established");
-
-        cron.schedule(CRON_INTERVAL_STRING, () => {
-            const now = new Date().toLocaleTimeString();
-            socket.emit(DELIVER_NOTIFICATION, `Cron job fired at ${now}`);
+        cron.schedule(CRON_INTERVAL_STRING, async () => {
+            await fireNotifications(socket);
         });
 
         socket.on(DISCONNECT, () => {
             console.log("notiSched utils: connection terminated");
         });
     });
+};
+
+// Gets users to notify and pushes notifications to them
+const fireNotifications = async (socket) => {
+    const now = new Date();
+    const noNotificationsMessage = `No one to notify at ${now.toLocaleTimeString()}`;
+    const usersToNotify = await getUsersToNotify(now);
+
+    usersToNotify.length ? notifyUsers(usersToNotify, socket) : socket.emit(DELIVER_NOTIFICATION, noNotificationsMessage);
+};
+
+// Calculates users' best notification times and queues them to be notified if it falls within the cron job window
+const getUsersToNotify = async (now) => {
+    const nowAsSecondOfDay = toSecondOfDay(now);
+    const halfHourBefore = nowAsSecondOfDay - HALF_HOUR_IN_SECONDS;
+    const halfHourAfter = nowAsSecondOfDay + HALF_HOUR_IN_SECONDS;
+
+    const users = await prisma.user.findMany();
+
+    // Iterate users and calculate best times
+    for (const user of users) {
+        delete user.password;
+        const sessionData = await prisma.sessionData.findUnique({ where: { userID: user.userID } });
+        // const interactionData = await prisma.interactionData.find({ where: { userID: user.userID } });
+
+        // Default user's best time to midday if they have no session logged
+        if (sessionData.sessionCount === 0) {
+            user["bestTime"] = MIDDLE_OF_DAY_IN_SECONDS;
+        } else {
+            user["bestTime"] = Math.ceil(sessionData.averageSessionStartTime + sessionData.averageSessionTime / 2);
+        }
+    }
+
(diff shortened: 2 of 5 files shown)


server/utils/notificationSchedulingUtils.js
```diff
@@ -1,19 +1,83 @@
...
+const getUsersToNotify = async (now) => {
+    const nowAsSecondOfDay = toSecondOfDay(now);
+    const halfHourBefore = nowAsSecondOfDay - HALF_HOUR_IN_SECONDS;
+    const halfHourAfter = nowAsSecondOfDay + HALF_HOUR_IN_SECONDS;
+
+    const users = await prisma.user.findMany();
+
+    // Iterate users and calculate best times
+    for (const user of users) {
+        delete user.password;
```
chris4307, line 48: it might be good to have a user data fetching function to be used across the app, so you don't forget to delete the password

server/utils/notificationSchedulingUtils.js
```diff
@@ -1,19 +1,83 @@
...
+    return notifiableUsers;
+};
+
+// Grab top candidate for each user lined up and emit the notification through the WebSocket
+const notifyUsers = async (usersToNotify, socket) => {
+    for (const user of usersToNotify) {
+        let topCandidate;
+
+        // Acquire candidates and pick out top candidate (first in array)
+        const res = await fetch(`http://localhost:3000/recommendations/acquireCandidates/for/${user.userID}`).catch((error) => {
```
chris4307, line 74: is this supposed to be a localhost request?

server/utils/notificationSchedulingUtils.js
```diff
@@ -1,19 +1,83 @@
...
+            console.log(error);
+        });
+        const candidates = await res.json();
+        topCandidate = candidates[0];
```
chris4307, line 78: it'd be interesting if you implemented an endpoint that only returns the single top recommended candidate (possibly filtered for users that have recently been recommended already). Would reduce the network load since you never use the other candidates anyway

client/src/utils/clientWebSocketUtils.js
```diff
@@ -8,9 +8,7 @@ export const establishWebSocketConnection = (setNotifications, setHasNewNotifica
     const identifier = new Date().getMilliseconds();
 
     const socket = io(VITE_SERVER_URL);
-    socket.on(CONNECT, () => {
-        console.log(`Websocket connection established at ${new Date().toLocaleTimeString()}`);
-    });
+    socket.on(CONNECT, () => {});
```
chris4307, line 11: just curious, and sorry if I missed it if you explained somewhere already, but why did you elect to use websockets vs. HTTP? In isolation, I wonder if a persistent connection is necessary for this functionality. 

HTTP implementation would be quite different (you'd have to poll from the client when it's time to fetch a notification, vs. a cron job that pushes a notification to the client). 

Curious to know any tradeoffs you considered here!

chris4307 (commented):
very cool!

server/utils/notificationSchedulingUtils.js
```diff
@@ -1,19 +1,83 @@
...
+    for (const user of users) {
+        delete user.password;
+        const sessionData = await prisma.sessionData.findUnique({ where: { userID: user.userID } });
+        // const interactionData = await prisma.interactionData.find({ where: { userID: user.userID } });
+
+        // Default user's best time to midday if they have no session logged
+        if (sessionData.sessionCount === 0) {
+            user["bestTime"] = MIDDLE_OF_DAY_IN_SECONDS;
+        } else {
+            user["bestTime"] = Math.ceil(sessionData.averageSessionStartTime + sessionData.averageSessionTime / 2);
```
stephanieprussick, line 56: just want to make sure this is the formula you want
PEMDAS applies in code :) 

if you want 
(sessionData.averageSessionStartTime + sessionData.averageSessionTime)/2 instead

server/utils/notificationSchedulingUtils.js
```diff
@@ -1,19 +1,83 @@
...
+        return u.bestTime >= halfHourBefore && u.bestTime < halfHourAfter;
+    });
+
+    return notifiableUsers;
+};
+
+// Grab top candidate for each user lined up and emit the notification through the WebSocket
+const notifyUsers = async (usersToNotify, socket) => {
+    for (const user of usersToNotify) {
+        let topCandidate;
```
stephanieprussick, line 71: why do you need this let here? doesn't seem like you are setting topCandidate more than once within the block?

server/utils/notificationSchedulingUtils.js
```diff
@@ -1,19 +1,83 @@
...
+    const halfHourBefore = nowAsSecondOfDay - HALF_HOUR_IN_SECONDS;
+    const halfHourAfter = nowAsSecondOfDay + HALF_HOUR_IN_SECONDS;
+
+    const users = await prisma.user.findMany();
+
+    // Iterate users and calculate best times
+    for (const user of users) {
+        delete user.password;
+        const sessionData = await prisma.sessionData.findUnique({ where: { userID: user.userID } });
+        // const interactionData = await prisma.interactionData.find({ where: { userID: user.userID } });
```
stephanieprussick, line 50: nit - delete this comment?

server/utils/notificationSchedulingUtils.js
z3nL: I'm not sure I understand - do you mean if it was supposed to be a url env import that I _could_ make local host / whatever URL I use in deployment?

server/utils/notificationSchedulingUtils.js
z3nL: Def puts a refactor into question because there's no situation where I use the rest of the candidates besides just showing that the algo. works in test plans

client/src/utils/clientWebSocketUtils.js
z3nL: A persistent connection isn't absolutely necessary - I was originally gonna go with HTTP long polling because it could've served my case pretty well, but I thought push-based delivery with WebSockets would be fun to try.

While it's out of the scope of relevancy for my app, the additional overhead to establish a connection every time a notification would need to be fetched seemed less intuitive than establishing a continuous, bi-directional one, the latter of which aligns with my sentiment of wanting to minimize the work the client itself needs to do to retrieve information (i.e. just connect and let the cron job in the server handle the rest).

Also, the Socket.io library just falls back on HTTP if WebSocket fails, so that did have bearing on my decision in the sense that my original intent would still be present in implementation  🙂‍↕️

server/utils/notificationSchedulingUtils.js
z3nL: PEMDAS errors are not lost on me to this day, but I did mean to only divide averageSessionTime by 2 to make the best time to send a notification be the middle of a user's session relative to their average start time. Adding a comment describing this to make that line more readable

server/utils/notificationSchedulingUtils.js
z3nL: You're right, I don't - I've removed it since this PR

server/utils/notificationSchedulingUtils.js
z3nL: Fair enough, it was mostly there as a reminder of some plans I had in mind for a better way to decide best time - more on that soon if I go through with it

PR #38, opened 2025-07-23
[TC2] Multiple Clients Receive Personalized Notifications Using Socket Rooms

z3nL:
## Preface
#37 implemented firing follow suggestion notifications at regular intervals for users whose "best times" fell in the allowance window of the cron job. The caveat to this, however, was that the test plan was deliberately structured so that only the _one_ user that was signed in fell within the allowance window during the test so we could easily verify what happens when that user falls out ("nothing", expectedly). 

## Overview
The notification system has been expanded so that multiple users can be signed into the app at the same time and receive follow suggestions personalized to them.
- On connection to the WebSocket server after user sign-in, the client socket instance emits an event to subscribe the server socket instance to a private room marked by the user's userID, stylized "user_ _userID"_

- The cron job runs the same as previously implemented, eventually calling **notifyUsers** if needed. 

- Instead of globally emitting a notification event on completion, **notifyUsers** now emits a notification event to a user through their corresponding channel name.

- If that user is signed in to the application, then their client socket instance will be subscribed to the channel and listening for the notification event, after which it will extract its content and display it in the notifications popup

### Two-User Display Check
- Remove the best time allowance window so we are guaranteed that _all_ users will receive a notification on cron job occurrence
- Have two browser instances open, each signed into Sk8rlog on a different account to facilitate the test
- Wait for the cron job to run and create a notification
- Both users should receive a notification at approximately the same time, and the message will be personalized for them.

Video of successful run of outlined plan (video starts at AS the notification fires due to file size constraints)
https://github.com/user-attachments/assets/d770476e-697d-4090-9cb2-13ea67238121

### DIFF
--- a/server/utils/notificationSchedulingUtils.js
+++ b/server/utils/notificationSchedulingUtils.js
@@ -53,7 +53,7 @@ const getUsersToNotify = async (now) => {
 
 // Grab top candidate for each user lined up and emit the notification through the WebSocket
 const notifyUsers = async (usersToNotify, socketServer) => {
-    for (const user of usersToNotify) {
+    usersToNotify.forEach(async (user) => {
         // Acquire candidates and pick out top candidate (first in array)
         const res = await fetch(`http://localhost:3000/recommendations/acquireCandidates/for/${user.userID}`).catch((error) => {
             console.log(error);
@@ -63,5 +63,5 @@ const notifyUsers = async (usersToNotify, socketServer) => {
 
         const suggestionMessage = `Hey ${user.name || `@${user.username}`}, you might be interested in @${topCandidate.username}`;
         socketServer.to(`user_${user.userID}`).emit(DELIVER_NOTIFICATION, `${suggestionMessage} | ${new Date().toLocaleTimeString()}`);
-    }
+    });
 };
--- a/server/utils/notificationSchedulingUtils.js
+++ b/server/utils/notificationSchedulingUtils.js
@@ -1,45 +1,31 @@
 import cron from "node-cron";
 import { toSecondOfDay } from "./sessionUtils.js";
-import {
-    CONNECTION,
-    DISCONNECT,
-    DELIVER_NOTIFICATION,
-    CRON_INTERVAL_STRING,
-    CRON_INTERVAL_DESCRIPTOR,
-    HALF_HOUR_IN_SECONDS,
-    MIDDLE_OF_DAY_IN_SECONDS,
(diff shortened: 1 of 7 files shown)


server/utils/profileRankingUtils.js
```diff
@@ -48,7 +48,7 @@ const evaluateCandidate = async (user, candidate) => {
     const cosineSimilarityScore = await computeSimilarityScore(user, candidate);
 
     // Convert the similarity score to a scalar factor
```
z3nL, line 50: Thank you @chris4307 and @stephanieprussick for the feedback on this file, the changes here are just pursuant to their comments/suggestions

server/utils/serverWebSocketUtils.js
```diff
@@ -1,18 +1,20 @@
 import { Server } from "socket.io";
-import { CONNECTION, DISCONNECT, REQUEST_NOTIFICATION } from "./constants.js";
+import { CONNECTION, DISCONNECT, REQUEST_NOTIFICATION, ENTER_ROOM } from "./constants.js";
 
 // Creates a websocket server and listens for notification requests
 export const createWebSocket = (server, corsConfig) => {
-    const socketServer = new Server(server, { cors: corsConfig });
+    const socketServer = new Server(server, { connectionStateRecovery: {}, cors: corsConfig });
```
z3nL, line 6: Default config to auto-attempt recovery when connection drops

server/utils/notificationSchedulingUtils.js
```diff
@@ -1,38 +1,24 @@
...
-    MIDDLE_OF_DAY_IN_SECONDS,
-} from "./constants.js";
+import { DELIVER_NOTIFICATION, CRON_INTERVAL_STRING, CRON_INTERVAL_DESCRIPTOR, HALF_HOUR_IN_SECONDS, MIDDLE_OF_DAY_IN_SECONDS } from "./constants.js";
 import { PrismaClient } from "../generated/prisma/index.js";
 const prisma = new PrismaClient();
 
 export const requestNotificationScheduling = (socketServer) => {
     console.log(`Cron generated - running a task every ${CRON_INTERVAL_DESCRIPTOR}`);
 
-    socketServer.on(CONNECTION, (socket) => {
```
z3nL, line 18: Every single time a client connected it would start a new cron job, misinterpreted the event listeners lol

z3nL (commented):
gen comments

PR #39, opened 2025-07-23
[TC2] Tracking a User's Suggested Users To Follow Across Operations

z3nL:
## Overview (Largely ported from [TC2 documentation](https://docs.google.com/document/d/1kNEz32Xm4RZwRLB867fBea2WWoBEhev3rD0eiksIfTs/edit?tab=t.0#heading=h.wxdbm3cim1iu))
1) Added new _suggestedUsers_ column in User model
-- Contains _Object_ { _String_ userID: _Date_ lastSuggestedDate , … }
-- Set of suggested users mapped to date they were suggested

2) When we suggest a user in **notifyUsers**,  we now update the user’s _suggestedUsers_ column to append a new suggestion instance

3) In the **evaluateCandidate** function:
-- Implemented a two day timeout for recently suggested candidates
-- If a user’s _suggestedUsers_ contains the candidate and it has been less than two days since _lastSuggestedDate_, skip evaluation by returning _-Inf_ as score
-- If it’s been longer than two days, release the candidate from timeout by removing them from user's _suggestedUsers_ in DB and proceed with evaluation

4) [MISC] When a user follows someone, remove them from the user's _suggestedUsers_ if they're present in it

### Process of Elimination (POE) Live Test - [Video Confirming Success](https://www.loom.com/share/4658ff09294f41018b1495a17cecf5ac?sid=de735554-e3cb-42fe-91bd-ff5c123b6652)
- Remove the best time allowance window so we are guaranteed a notification firing on cron job occurrence
- Add a test socket message that fires to the client when there are no users to display (i.e. every other user has already been suggested recently)
- Initialize an instance of the client by signing in. If all logic is sound, then the user will be suggested someone new each job occurrence, eventually receiving the fallback test socket message when all users have been suggested

### Post-POE Follow Test For Item 4
- Fired a message to server console if user follows someone who is in their _suggestedUsers_, prompting a removal and update in DB
image

### DIFF
--- a/server/routes/users.js
+++ b/server/routes/users.js
@@ -158,21 +158,26 @@ router.put("/:userID/followedUsers/:action", async (req, res, _next) => {
 
         // Remove user to follow from suggested users if they're present in it
         const recentlySuggestedUsers = user.suggestedUsers;
+        let isSuggestedUsersUpdated = false;
+
         if (recentlySuggestedUsers[userBeingReferencedID]) {
-            delete recentlySuggestedUsers[candidate.userID];
+            delete recentlySuggestedUsers[userBeingReferencedID];
+            isSuggestedUsersUpdated = true;
         }
 
-        await prisma.user.update({
-            where: { userID },
-            data: {
-                followedUsers:
-                    action === FOLLOW
-                        ? [...user.followedUsers, userBeingReferencedID]
-                        : user.followedUsers.filter((uID) => uID !== userBeingReferencedID),
+        // Only fire a DB query if we need to update the user's suggestedUsers
+        isSuggestedUsersUpdated &&
+            (await prisma.user.update({
+                where: { userID },
+                data: {
+                    followedUsers:
+                        action === FOLLOW
+                            ? [...user.followedUsers, userBeingReferencedID]
+                            : user.followedUsers.filter((uID) => uID !== userBeingReferencedID),
 
-                suggestedUsers: recentlySuggestedUsers,
-            },
-        });
+                    suggestedUsers: recentlySuggestedUsers,
+                },
+            }));
 
         return res.status(STATUS_CODES.OK).json({ message: "Followed users updated" });
     } catch (error) {
--- a/server/routes/users.js
+++ b/server/routes/users.js
@@ -28,6 +28,7 @@ router.post("/register", async (req, res) => {
                 password,
                 name,
                 location,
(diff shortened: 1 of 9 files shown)


z3nL (commented):
gen comments / q's if needed

server/routes/users.js
```diff
@@ -154,15 +155,29 @@ router.put("/:userID/followedUsers/:action", async (req, res, _next) => {
...
+                where: { userID },
+                data: {
+                    followedUsers:
+                        action === FOLLOW
+                            ? [...user.followedUsers, userBeingReferencedID]
+                            : user.followedUsers.filter((uID) => uID !== userBeingReferencedID),
+
+                    suggestedUsers: recentlySuggestedUsers,
+                },
+            }));
```
itumejia, line 180: I think for this logic that you are adding here, we could extract it to an util function to keep the route function simple?

server/utils/constants.js
```diff
@@ -59,3 +59,5 @@ export const UNFOLLOW = "unfollow";
 export const POST_OVR_WEIGHT = 0.7;
 export const POPULARITY_OVR_WEIGHT = 0.3;
 export const NO_CORRELATION = 0.0;
+
+export const RECENCY_CUTOFF_IN_DAYS = 2;
```
itumejia, line 63: constants files are getting quite long and it's a bit confusing since they are used across all the app. Couple ways to improve this:
- Easiest one: divide the file in sections by add comments to know in which part of the app they are being used (e.g. profile page, main feed page, posts ranking algo, profile ranking, etc. )
- Have multiple constant files. You can either divide the constants by components as mentioned in the previous bullet point, or even purpose (for example, having a file "content strings" that has strings that will be shown in the UI, and then having separate files that have more constants that are used for logical operations

You can also mix both tips!

server/routes/users.js
z3nL: Fair enough, I plead the fifth on whatever else might've been wrong here 😭 
[[hotfix] crit follow user error](https://github.com/Zen-L-MU2025/project-sk8rlog/commit/dc6a8546933ddbd10bc3fed7ad989d001841a46b)

server/utils/constants.js
z3nL: Ah, I thought constants was excused from proper organization/documentation, I'll handle this tonight 👍

PR #40, opened 2025-07-23
[TC2][Stretch] Iteration on Profile Ranking - New Metrics to Consider

z3nL:
## Context - Stretch Goal Documentation
Document ["Sk8rlog Planning" ](https://docs.google.com/document/d/1w-cblWeSteRe4uhV08d3gxD7hggLu-YP3gZbIwz27wk/edit?usp=sharing)→ Stretch Goals → [More Metrics](https://docs.google.com/document/d/1w-cblWeSteRe4uhV08d3gxD7hggLu-YP3gZbIwz27wk/edit?tab=t.0#heading=h.gmpabj78fy1w) → [TC2] Additional considerations when ranking a profile

## Overview
Implemented additional factors in the profile ranking algorithm pursuant to ideas discussed in documentation. New considerations that have been translated to numerical values include:
- A) A user does not follow the candidate but interacts with their posts in a meaningful capacity
- B) The user and candidate use the app at similar times

See link to documentation for details of implementation

### Process of Elimination (POE) Live Test - [Video Confirming Success](https://www.loom.com/share/e8884d572a154a20b45f10879ca26356?sid=d27f7db5-e580-4b84-a85d-9b739a09dc76)
Remove the best time allowance window so we are guaranteed a notification firing on cron job occurrence

Add a test socket message that fires to the client when there are no users to display (i.e. every other user has already been suggested recently)

Initialize an instance of the client by signing in to some **user**. If all logic is sound, then
* the **user** will be suggested someone new each job occurrence
* the server console will log generated total interaction points / meaningful interaction bonus (A) and user timing bonus (B) for each candidate as applicable and these values will show variation in some capacity across some candidates
* the client will eventually receive the fallback test socket message when all users have been suggested

### DIFF
--- a/server/utils/profileRankingUtils.js
+++ b/server/utils/profileRankingUtils.js
@@ -1,6 +1,20 @@
 import { PrismaClient } from "../generated/prisma/index.js";
-import { RANKING_MODES, POST_OVR_WEIGHT, POPULARITY_OVR_WEIGHT, NO_CORRELATION, MILLISECONDS_IN_DAY, RECENCY_CUTOFF_IN_DAYS } from "./constants.js";
+import {
+    RANKING_MODES,
+    POST_OVR_WEIGHT,
+    POPULARITY_OVR_WEIGHT,
+    NO_CORRELATION,
+    MILLISECONDS_IN_DAY,
+    RECENCY_CUTOFF_IN_DAYS,
+    RELEVANCY_CUTOFF,
+    MEANINFUL_INTERACTION_BONUS,
+    LIKE_POINT_VALUE,
+    COMMENT_POINT_VALUE,
+    INTERACTION_POINTS_THRESHOLD,
+    USER_TIMING_BONUS,
+} from "./constants.js";
 import { scorePosts } from "./serverPostRecommendationUtils.js";
+import similarity from "compute-cosine-similarity";
 
 const prisma = new PrismaClient();
 
@@ -11,9 +25,7 @@ export const rankCandidates = async (hostUserID) => {
     delete user.password;
 
     let candidates = await prisma.user.findMany({
-        // for when we have a following list
-        //where: { userID: { notIn: user.followedUsers } },
-        where: { userID: { not: hostUserID } },
+        where: { userID: { not: hostUserID, notIn: user.followedUsers } },
     });
 
     for (const candidate of candidates) {
@@ -27,7 +39,7 @@ export const rankCandidates = async (hostUserID) => {
     return candidates;
 };
 
-// Evaluate a suggestion candidate for a user based on their posts score, interests similarity, and mutual following
+// Evaluate a suggestion candidate for a user based on outlined metrics
 const evaluateCandidate = async (user, candidate) => {
     if (!user || !candidate) {
         console.log("evaluateCandidate: user or candidate is null");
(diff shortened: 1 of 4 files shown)


server/utils/profileRankingUtils.js
```diff
@@ -1,6 +1,20 @@
...
+    RECENCY_CUTOFF_IN_DAYS,
+    RELEVANCY_CUTOFF,
+    MEANINFUL_INTERACTION_BONUS,
+    LIKE_POINT_VALUE,
+    COMMENT_POINT_VALUE,
+    INTERACTION_POINTS_THRESHOLD,
+    USER_TIMING_BONUS,
+} from "./constants.js";
 import { scorePosts } from "./serverPostRecommendationUtils.js";
+import similarity from "compute-cosine-similarity";
```
z3nL, line 17: fun fact, my TC2 documentation says this package is broken because it didn't work the last time I tried to use it, turns out I wasn't supposed to wrap 'similarity' in curly braces

server/utils/profileRankingUtils.js
```diff
@@ -11,9 +25,7 @@ export const rankCandidates = async (hostUserID) => {
     delete user.password;
 
     let candidates = await prisma.user.findMany({
-        // for when we have a following list
-        //where: { userID: { notIn: user.followedUsers } },
-        where: { userID: { not: hostUserID } },
+        where: { userID: { not: hostUserID, notIn: user.followedUsers } },
```
z3nL, line 28: Now's a good time to invoke the already following restriction

server/utils/profileRankingUtils.js
```diff
@@ -65,21 +84,21 @@ const checkSuggestionsEligibility = async (user, candidate) => {
...
-            return false;
-        }
-
-        // It's been long enough, remove the candidate from the recently suggested list
-        else {
-            delete recentlySuggestedUsers[candidate.userID];
-            await prisma.user.update({ where: { userID: user.userID }, data: { suggestedUsers: recentlySuggestedUsers } });
-            return true;
-        }
+    if (!lastSuggestedDate) return true;
```
z3nL, line 87: This was actually a logic error before now since nothing would happen if lastSuggestedDate is not applicable

server/utils/profileRankingUtils.js
```diff
@@ -150,3 +174,76 @@ const computeMutualFollowingFrequency = (user, candidate) => {
...
+};
+
+// Generates a user's app usage vector based on session and interaction metrics, returns vector as array
+const createAppUsageVector = async (user) => {
+    const userSessionData = await prisma.sessionData.findUnique({ where: { userID: user.userID } });
+    const userInteractionData = await prisma.interactionData.findUnique({ where: { userID: user.userID } });
+
+    const userAverageHighInteractionDensityTime = computeAverageHighInteractionDensityTime(userInteractionData);
+
+    // Abbreviate to reduce clutter
```
z3nL, line 221: Is abbreviating a var on the spot like this permissible or should I stay away from it? My thought process was that anyone reading up to this point can clearly see the original name so shortening it for the calculation can safely reduce the number of characters on the screen

server/utils/profileRankingUtils.js
```diff
@@ -150,3 +174,76 @@ const computeMutualFollowingFrequency = (user, candidate) => {
...
+    const u_S_D = userSessionData;
+    const userAppUsageVector = [u_S_D.averageSessionStartTime, u_S_D.averageSessionTime, userAverageHighInteractionDensityTime];
+
+    return userAppUsageVector;
+};
+
+// Returns average high interaction density time for provided user interaction data
+const computeAverageHighInteractionDensityTime = (interactionData) => {
+    // Abbreviate in these calculations to mitigate clutter
```
z3nL, line 230: See prev comment

z3nL (commented):
gen comments / q's

server/utils/profileRankingUtils.js
itumejia, line 1: I've noticed in your latest PRs, that your util files are getting quite long. This is natural as you app keeps growing, but it is also a sign that it may be a time for a refactor :)

See my [comment](https://github.com/Zen-L-MU2025/project-sk8rlog/pull/39#discussion_r2227039028) I did for constants files, it applies here in a similar way. Few things I would advice:
- I think we can start having multiple directories for our util functions. I'm thinking we could at least have one for notifications, and one for each ranking algorithm. We'll need this as you'll probably end up having more files after the refactor, so we need to keep the entire codebase clean
- For each utils file (specially the ones that are getting large), think about how you can modularize this in a way that is easier to understand. For example, by looking at this file I think it makes sense to split it at least in 2 different files: one file with functions to calculate each individual metric, and one that will consolidate them and have all the logic to make a final evaluation

server/utils/profileRankingUtils.js
itumejia: I'd say no. We should make sure every line of code is easy to read. Although having "smaller" lines of code is a valid concern at some level, it is not enough to use abbreviations. 
Some ways to mitigate this are splitting a line of code in multiple statements if you are appending multiple functions calls, or, if the lines cannot be split into more statements, you can always split a single statement into multiple lines, specially if you are working with arrays or objects.

server/utils/profileRankingUtils.js
itumejia: Also this can get as granular as you want! For example, if a simple metric is getting complex enough that is taking a significant amount of code, you can extract that into a separate file too. This can also result in having files with single functionalities, instead of having a bunch of util files. 

For example, you could have a file named `calculateMeaningfulInteractionBonus.js` , which will essentially be exporting a single function (you can use [default exports](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export#:~:text=value%20of%20%60x%60.-,Default%20exports%3A,-JS) in this case). And as you will only be exporting one function, you can add as many private functions as you want in that file that will be solely be used to make that exported function work. This would be a huge code quality boost as it will be very easy to see which functions are used only to make an exported function operate correctly

server/utils/profileRankingUtils.js
```diff
@@ -88,20 +107,25 @@ const computeCandidatePostMetrics = async (user, candidate) => {
...
+    // Average post scores as an overall posts score
+    let rankedCandidatePosts = await scorePosts(candidatePosts, user, RANKING_MODES.RECOMMENDED);
+    let totalScore = 0;
+    let totalPopularity = 0;
+    for (const post of rankedCandidatePosts) {
+        totalScore += post.score;
+        totalPopularity += post.popularity;
     }
+    postOvr = totalScore / candidatePosts.length;
+    popularityOvr = totalPopularity / candidatePosts.length;
```
itumejia, line 126: speaking of abbreviations, let's deabbreviate these

PR #41, opened 2025-07-24
[TC2][Stretch] Improve Notifications Flow

z3nL:
### Overview
The user suggestion flow has been reworked to facilitate the incoming post notifications feature.

- In **notifyUsers**, the data the WebSocket server fires as a notification is now a more descriptive payload that consists of recipient as it should be displayed in client, notification type, data corresponding to the notification, and notification timestamp.

- In the client codebase, _notificationsUtils_ has been refactored into a JSX file (from JS) to establish a function **formatNotification** that destructures a notification payload and returns an HTML element that properly represents the notification's contents.

- When the _Header_ component maps the notifications array hook, it invokes **formatNotification** instead of returning the historic default `<p>` message element

### Test Plan
Ran the Process of Elimination Live Test (see usage examples in #39 and #40) and confirmed that clicking the suggested user successfully redirected to their profile
[Video confirming success](https://github.com/user-attachments/assets/a9c10e2e-eeb9-4e3f-8cda-93854884c88c)


server/utils/notificationSchedulingUtils.js
```diff
@@ -28,6 +35,7 @@ const getUsersToNotify = async (now) => {
     const halfHourAfter = nowAsSecondOfDay + HALF_HOUR_IN_SECONDS;
 
     const users = await prisma.user.findMany();
+    // TESTING: return users here to trigger mass notify
```
itumejia, line 38: let's remove this?

PR #43, opened 2025-07-25
[TC1][TC2][Stretch] Post Notifications

z3nL:
# Overview
Any time a post is created, its post score is generated for each user based on the TC1 post recommendation algorithm. 
For users whose generated score for the post exceeds a defined threshold, a notification is fired in real time that they might be interested in using the post using the notifications functionality created for TC2.
The notification has a clickable redirect to the post's individual page

### Notes
- Facilitated creation of the feature by extracting the functionality of scoring a single post in _scorePosts_ into its own file _calculateSinglePostScore_ as the default export
- A wrapper that initializes the parameters needed to call  _calculateSinglePostScore_ is layered on top of it and is what gets invoked by the post notification flow 
- Refined the formula for scoring a single post to be less critical of popularity, promoting better results
### File Refactors
- [Server] Renamed _getUsersToNotify_ as _getUsersToSendCandidateTo_
- [Server] Renamed _notifyUsers_ as _notifyUsersWithCandidate_
- [Server] Updated imports in _requestNotificationScheduling_ to reflect these changes

# Test Plan
### Full Notification Sequence Simulation - [Video Confirming Success](https://github.com/user-attachments/assets/7f645113-25ae-49e2-8a22-3bd2476fb73e)
- Create a post through user A's account that contains keywords **irrelevant** to user B but **relevant** to user C
- Wait for successful post creation
- Check user B's notifications and verify they are expectedly empty
- Check user C's notifications, verify there is expectedly a post suggestion notification, and verify the redirect to the post's individual page works

### DIFF
--- a/server/utils/postRecommendations/calculateSinglePostScore.js
+++ b/server/utils/postRecommendations/calculateSinglePostScore.js
@@ -24,7 +24,8 @@ const calculateSinglePostScore = async (
 ) => {
     let rawPostScore = 0;
 
-    const popularityScore = post.likeCount * LIKE_WEIGHT + post.comments?.length * COMMENT_WEIGHT;
+    // Default popularity score to 1
+    const popularityScore = 1 + post.likeCount * LIKE_WEIGHT + post.comments?.length * COMMENT_WEIGHT;
 
     const filteredPostContent = await filterTokens(post.description);
     let overlap = {};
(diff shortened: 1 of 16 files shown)


server/utils/postRecommendations/calculateSinglePostScore.js
```diff
@@ -0,0 +1,79 @@
+import { LIKE_WEIGHT, COMMENT_WEIGHT, CLIPS, NOT_APPLICABLE } from "../constants.js";
+import calculateBiasFactors from "./helpers/calculateBiasFactors.js";
+import filterTokens from "./helpers/filterTokens.js";
+import calculateTokenScore from "./helpers/calculateTokenScore.js";
+import getPostLength from "./helpers/getPostLength.js";
+
+// Wrapper function for calculateSinglePostScore > takes in a user obj as argument to generate full set of params
+// Used in the post notifications flow
+export const calculateSinglePostScoreWithUserObject = async (post, user) => {
```
itumejia, line 9: noob question - what's the difference between this and what we use in the raking algorithm to give a score to posts?

z3nL: There isn't

 _calculateSinglePostScore_ takes params:
- post,
- userFrequency,
- avgLengthOfLikedPosts,
- portionOfLikedPostsThatAreClips,
- portionOfLikedPostsThatAreBlogs,
- userLocation

In the posts ranking algorithm where we score ALL posts for a user, we acquire these vars at the start and repeatedly invoke _calculateSinglePostScore_, passing these vars as params each time

However, when we're scoring a single, newly created post, these vars aren't readily available because we didn't call the full scoring algorithm for all posts. _calculateSinglePostScoreWithUserObject_ takes the user we're scoring it for and generates the parameters necessary to invoke _calculateSinglePostScore_ just once for a user

PR #44, opened 2025-07-25
[TC1][TC2][Stretch] Selecting Location in Registration

z3nL:
### Overview
To facilitate the implementation of location metrics in post and profile ranking, new users now opt to select their location from a set dropdown menu in the registration modal instead of manually typing it because the latter generates too much overhead in translating location input to coordinates

### Test Plan
[Video verifying ](
https://github.com/user-attachments/assets/f9274375-7642-4e85-af7a-7591536d9dd8)that creating a user is still fully functional and their selected location is visible on their profile

### DIFF
--- a/client/src/utils/constants.js
+++ b/client/src/utils/constants.js
@@ -6,6 +6,16 @@ export const OPTIONS = {
     REGISTER: "register",
 };
 
+// Selectable locations on signup
+export const LOCATIONS = {
+    DEFAULT: "none",
+    MIA: "Miami",
+    SEA: "Seattle",
+    SFO: "San Francisco",
+    LAX: "Los Angeles",
+    NYC: "New York City",
+};
+
 // Post types to dictate views in app
 export const CLIPS = "Clips";
 export const BLOGS = "Blogs";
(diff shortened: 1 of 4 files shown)


client/src/utils/constants.js
```diff
@@ -67,7 +77,7 @@ export const NON_ALPHANUMERIC_REGEX = /[^a-zA-Z0-9]/;
 // For filtering posts
 export const RANKING_MODES = {
     RECOMMENDED: "reccommended",
-    LATEST: "latest",
+    LATEST: "latitudeest",
```
itumejia, line 80: what is this? lol

itumejia: not sure if I'm following correctly, but if you are replacing "latest" ranking with the one from location, I think we should keep both? and probably "closest" would be more descriptive?

z3nL: Disregard this change in specific, I unknowingly generated this change when running a mass replace from "lat" to "latitude" in the codebase and it caused a bug I had to mitigate just earlier in #46

PR #45, opened 2025-07-25
[TC1][TC2][Stretch] Implemented Locations as a Factor in Ranking

z3nL:
## Overview
Implemented a function that:
- maps user/post locations to coordinates and performs spherical distance calculations using the Haversine formula
- uses this distance in a rational square root function to generate a proximity bias factor applied to either a user-post evaluation or a user-candidate evaluation, effectively integrating this functionality into the post recommendation algorithm / notification system and profile ranking / notification system

## Test Plan
Leveraged the previously implemented post notifications feature (#43) to actively log proximity bias calculation results to the console when a new post gets evaluated for every user. All outputs are in alignment with intended results of calculations:

image

Bias factor is in play when automatically scoring posts for Home/Clips/Blogs pages
image

### DIFF
--- a/server/package-lock.json
+++ b/server/package-lock.json
@@ -20,6 +20,7 @@
         "bcryptjs": "^3.0.2",
         "compute-cosine-similarity": "^1.1.0",
         "get-video-duration": "^4.1.0",
+        "haversine": "^1.1.1",
         "jsonwebtoken": "^9.0.2",
         "multer": "^2.0.1",
         "node-cron": "^4.2.1",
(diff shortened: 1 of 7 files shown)


itumejia:
nit on the test plan: since you are also using this in the home feed ranking algorithm (correct me if I'm wrong), can you also add something that shows that it works well for ranking?

z3nL:
> nit on the test plan: since you are also using this in the home feed ranking algorithm (correct me if I'm wrong), can you also add something that shows that it works well for ranking?
@itumejia 

Yeah fair enough, just added it

